Issue #9, opened 2016-05-20
Minimap information on mouseOver


jfastnacht:
Add map name and coordinates when mouse over minimap.

jfastnacht:
Example:
![image](https://cloud.githubusercontent.com/assets/6183607/15431281/66c48680-1ea9-11e6-8441-c80a88157d52.png)

shagu:
Or like this:
http://shaguaddons.ericmauser.de/shaguui/img/mmbuff.png

Should be adjustable in config.

PR #12, opened 2016-05-20
Zone text and coordinates

jfastnacht:
Added zone text under minimap and coordinates in minimap on mouse hover.

image


jfastnacht:
Issue https://github.com/shagu/pfUI/issues/9

shagu:
merged

PR #38, opened 2016-10-09
Add note about bag panel to FAQs

muellerj:


shagu:
Thanks for your efforts, but I did a complete rewrite of the bag and bankframes today. (fb323e2)
Now there's a dedicated button for that :)

PR #60, opened 2016-10-26
Easy config for changing chat height

Ko0z:
After every update I have to manually change the left and right chat height to 200 from 150, I thought it would be easier if we could have an option for this in the Settings UI. Thanks!


shagu:
https://github.com/shagu/pfUI/commit/4d51cf8bb17c5c4859d3e5f8ce73254f77674afd

PR #69, opened 2016-11-07
Added bag sorting

shirsig:
Retail blizzlike sort button + sound in the bag frame.


shagu:
Am I allowed to refactor this commit?
I'd like to exchange the icon, remove the sound and make a pfUI-module out of it.
Thanks for your efforts, this is an awesome piece of code! :)

shirsig:
Well, I'd prefer if you kept the code within the library mostly intact so I can keep it synchronized with my repo at https://github.com/shirsig/Clean_Up-lib. Don't care about the icon/sound.

shagu:
I see... Then I'd rather add it as a thirdparty addon instead.
Different code styles and code practices with various different folders of libraries and modules brought ShaguUI to an unmaintanable piece of sh*t. I don't want to have these kind of problems again. This as the only one, wouldn't be that bad, but If I accept this one, others will refer to it and want theirs "integrated" the same way, which would lead to the same troubles. 
I'm sorry and I hope you understand my concerns.

shirsig:
Hm, well, it only exposes a single function, don't think that would cause maintainability issues.
But anyway, I guess that works too. I've added a .toc for standalone use (https://github.com/shirsig/Clean_Up-lib). Maybe you could add a button to the bag/bank when it's enabled similar to your WIM support? The API is the same.

shagu:
> Maybe you could add a button to the bag/bank when it's enabled similar to your WIM support?
That's exactly what I mean with "thirdparty" integration. So you will keep being the maintainer of your code. Roughly spoken, in thirdparty.lua, it would just be like this:
```
if Clean_Up then
 prepare everything
 create buttons
 and so on
end
```
I'll do it this evening if I find some spare time. Thanks for your contribution.

Cheers

shirsig:
Great, thanks :)

shagu:
I'd like to do something like this, to sort the bank (if bankframe is opened):
```
     pfUI.bag.right.sort:SetScript("OnClick", function()
       Clean_Up("bags")
       if pfUI.bag.left:IsShown() then
         Clean_Up("bank")
       end
     end)
```
But there seems to be an Issue by calling the function twice in a short time. 
Would be nice if you could add something like Clean_Up("all").

Cheers

shagu:
https://github.com/shagu/pfUI/commit/446cf795b011e4293914e93f876c6c6621cec080

shirsig:
Dunno, isn't it better to have a separate button? Some may not want to sort one of bag/bank at all, and you also might have some stuff manually placed for trading/crafting that you don't want messed up when sorting the bank. Doing both at the same time is actually a non trivial change (unless the sorting would be across the two, which I don't think is what you meant)

shagu:
https://github.com/shagu/pfUI/issues/74

PR #123, opened 2016-11-24
Added option to hide right chat background.

stokbaek:
I made a small hack to be able to disable the background of the right panel window. This is just a hack it should properly be handled different but it works for now.


Cruzix:
Is this only the right chat background? Cuz I'd still want the panel to be there for my personal ui ^^

stokbaek:
Looked abit more through the code. There is a button to hide the chat part. I found a better way to hide it which I will push soon. This will hide it all, also the button to toggle the chat panel

Cruzix:
Aw okay. ye I know theres a button on the panel that hides the right chat background (If you dont have addons embedded through the thirdparty option ingame) but it dosnt "save" on logout/reload so having to  manually hide it every time is a bit annoying

stokbaek:
I will cancel this PR as I have started to make an addon that changes some settings in pfUI to make it more minimalistic. BTW thanks for this great vanilla wow overhaul

PR #167, opened 2016-12-18
thirdparty: DPSMate integration

Road-block:
**pfUI.thirdparty: Fix issues with loading order breaking DPSMate integration.**  
DPSMate has not always finished loading when pfUI tries to configure and panel-ize it.  
DPSMate UI is initialized after it loads its sv.


shagu:
https://github.com/shagu/pfUI/commit/8490f49c4cd208ef4fb2103c1ac42a3d2d6c6b6a

Ekimin:
Hi shagu,
I found that pfui does not support Chinese characters well.(My client is original English Client with Chinese fonts).  I changed the fonts in pfui and it works well now.

Issue #168, opened 2016-12-19
Add locale compatible font (zhCN)


Ekimin:
Hi there,
Thx for ur useful addon it helps me a lot.
There's 2 improvements which I found could be done:
1.The default setting of chatbox width and height are 0,which causes the problem that input characters couldn't be seen while typing.(30 height and 200 width is ok)
2.the fonts in the addon do not support Chinese characters in English Client. I use English Client with Chinese fonts, and I found i cant see any Chinese characters using pfui. I changed the fonts(arial) in your addon with Chinese fonts and it works well with both English and Chinese characters.

shagu:
Hi,

1. the width and height are set to 0 but this actually means that pfUI is using its builtin settings, which are 22 height and actionbar width (https://github.com/shagu/pfUI/blob/master/modules/chat.lua#L469-L474)

2. Is there a arial alike font which also provides chinese characters somewhere? If yes, could you please link it?

Ekimin:
Hi,
1. I found that my addon do not show what i type if i do not set width and height manually.Maybe it is a individual case for me.
2. I'v post a pull request in which i added a arial alike font. It's unicode fonts so it  provides both Chinese characters and English ones too.

Ekimin:
Here is the font link(arial unicode ms.ttf):  https://github.com/Ekimin/pfUI/tree/master/fonts

shagu:
Wow! It's 23 MB. I won't add that. Sorry.

shagu:
Does the Wow default font (without pfUI) support chinese characters?

shirsig:
The fonts shipped with the english client don't, but I assume the default font (as assigned to STANDARD_TEXT_FONT) varies by client and if you use that it would work for every client.

shagu:
As a first step, this adds support for zhCN gameclients: https://github.com/shagu/pfUI/commit/77f02d96743bc2ede861c697ca8c4e62578e8448

chinese characters are still not visible on non chinese gameclients. 
(because the font is not available for non chinese clients)

jokers-x:
thinks for you very nice addon , but the pfUI addons nameplates or ShaguPlates can't work on cwow

Ekimin:
> chinese characters are still not visible on non chinese gameclients.
(because the font is not available for non chinese clients)
----
[...]

PR #169, opened 2016-12-19
Added new font to support Chinese input.

Ekimin:
Hi, here i added the font file in this pull request.
>Here is the issue link:     https://github.com/shagu/pfUI/issues/168


shagu:
Thanks for your effort, but 23 MB is way to much. I'll keep on searching for an appropriate font.

PR #220, opened 2016-12-23
Fixed bug with eqcompare frames

Frizlee:
When you mouse over item, press shift and then release it frame didn't hide. I made a simple bugfix for that.


shagu:
Thanks for your report, but I won't merge this:
ShoopingTooltip is a Tooltip which is used by the default UI. If we going to hide this everytime when no ShiftKey is pressed, we're definitly going to break other things.

PR #225, opened 2016-12-24
De-globalizing some utility functions

Road-block:
I found a conflict with the addon [RangeHelp](https://wow.curseforge.com/projects/project-880/files/64666) by accident; _my brother plays a Hunter and I'm maintaining his UI_.  
It loads after pfUI and overwrites the `CopyTable` function with another version (worse) that breaks pfUI profile functionality.  
One option would be to make the global name more unique (for example `pfUICopyTable`) but namespacing methods used by the addon (wrapping the function inside the pfUI object) is better solution imho.  
Since `strsplit` and `round` are also quite common variable names I did the same.  
I also made a cosmetic change: Closing open dialogs if the main config frame closes (question, colorframe).


shagu:
I did it my way. Anyway thanks for the report and your efforts. Made this as a first step, with a vision in mind of how my api should look like in future. https://github.com/shagu/pfUI/commit/ad4e68d681d8e06a888b2026eebccf5caeeeb5a7

shagu:
rebased: https://github.com/shagu/pfUI/commit/848ce1d700d245f1b3ca109e7d212c6474fae17e

Issue #226, opened 2016-12-24
Equipment comparison window not cleaned up properly


torusnc:
The recently implemented equipment comparison window becomes stuck on screen, moving with the tooltip (currently, tooltip is set to 'cursor' in my options). Each item that regenerates a tooltip moves the item window to that location, as if the comparison is continuing. Please see screenshots below at the following link:

https://imgur.com/a/6jW33

Thank you!

Barry031:
This bug is caused by the addon Questie, revert back to 3.2 or wait for the questie dev to update it.
Working fine for me with quetie disabled.

Barry031:
This is a questie addon bug, disable it or revert back to an older version. Working fine for me with latest pfUI build and questie disabled.

torusnc:
Great, thanks for the note! I'll raise the issue on Questie's github.

torusnc:
Questie has been updated to address this issue, tests fine on my end.

PR #238, opened 2016-12-27
eqcompare: shopping tooltip not hiding bug fixed

Frizlee:
Next approach to fix this bug. Now it works pretty well and doesn't mess with ShoppingTooltip scripts.


shagu:
Is it because of this bug https://github.com/shagu/pfUI/issues/226 ?

Frizlee:
Hmm in fact I'm using newest questie. I reverted changes and disabled it and bug is gone.

shagu:
I'll have a look later.

shagu:
https://github.com/AeroScripts/QuestieDev/issues/310

PR #242, opened 2016-12-28
Making Actionbars more configurable

Road-block:
- Fixes durability panel tooltip not refreshing (closures don't recalculate up-values)
- Decouples Actionbar1/BonusBar from other bars so they can all be moved and managed separately.
- Make Bar Layouts possible (any valid combination of columns by rows)
- Add configuration options and defaults for bar layouts (Actionbar1 by default on top).
- Vertical Labels for mover when needed.
- Using appropriate globals where available instead of hardcoded values.
- Various pfUI.api additions.

Check out the short clip below to get an idea what it does (set quality to highest available).
__I still have one major feature to add before I've done what I had in mind.__
Let it sit or close it if you like, 😃  
I'll update the pull request when I've finished and you can use or discard as you prefer (no pressure 😋 )
[Video](https://vid.me/st7X)


shagu:
Hi, Would it be possible to rebase your work to the latest "master" and create a clean PR instead?
Thanks in advance.

Issue #418, opened 2017-02-03
Extendable OnClick support for Unit/Raid frames


kxseven:
Would you consider replacing the OnClick function created separately in each unit frame (player/target/party/raid) with a generic one that accepted `button` and `unitid`?

```lua
function pfUI_OnClick_Handler(button, unitid)

    -- use the custom OnClick function if one is hooked
    if (type(pfUI_Custom_ClickFunction) == "function") then
        if (pfUI_Custom_ClickFunction(button, unitid)) then
            -- pfUI_Custom_ClickFunction should return true if handled, then we do nothing more.
            -- no return and we'll continue with default pfUI behaviour
            return true
        end
    end

    -- existing pfUI OnClick code/behaviour
    ...

end
```

 - This would allow other AddOns (healing orientated ones primarily) to hook their custom function. 
 - This is how it was coded in PerlClassic and X-Perl which then allowed support for AddOns such as Clique and HealBot
 - I've prepared a version of [Clique](https://github.com/kxseven/Clique/commit/acc9c51a75a496901b3d1f550b4879c7441fd54c) with support for pfUI if this change is made.

**Related Issues:**
https://github.com/shagu/pfUI/issues/363
https://github.com/shagu/pfUI/issues/104

kxseven:
Updating;
 - I've started work on a fork of pfUI with support added for player, target and targettarget frames.
 - I'll submit a PR once I've got party and raid frames working also.

kxseven:
PR submitted https://github.com/shagu/pfUI/pull/419

shagu:
Thanks.

kxseven:
Updated PR https://github.com/shagu/pfUI/pull/436 based on current implementation of OnClick function

shagu:
https://github.com/shagu/Clique/commit/a5ee56c3f803afbdda07bae9cd330e0d4a75d75a

PR #419, opened 2017-02-03
Initial changes for custom OnClick support

kxseven:
Ref: https://github.com/shagu/pfUI/issues/418

Summary of changes:
 - Moved all of the per-module OnClick() functions into a generic pfUI_OnClick_Handler()
 - Replaced existing OnClick() calls calls to pfUI_OnClick_Handler()
 - Registered frames for all 5 mouse buttons (instead of just LeftClick, RightClick)

Tested and working with updated version of [Clique](https://github.com/kxseven/Clique) that has pfUI support

### DIFF
--- a/modules/group.lua
+++ b/modules/group.lua
@@ -153,7 +153,7 @@ pfUI:RegisterModule("group", function ()
     pfUI.uf.group[i].hp.raidIcon:SetPoint("TOP", pfUI.uf.group[i].hp, "TOP", -4, 4)
     pfUI.uf.group[i].hp.raidIcon:Hide()
 
-    pfUI.uf.group[i]:RegisterForClicks('LeftButtonUp', 'RightButtonUp')
+    pfUI.uf.group[i]:RegisterForClicks('LeftButtonUp', 'RightButtonUp', 'MiddleButtonUp', 'Button4Up', 'Button5Up')
 
     pfUI.uf:CreateUnit(pfUI.uf.group[i])
   end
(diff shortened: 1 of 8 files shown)


shagu:
https://github.com/shagu/pfUI/commit/c0d33cd02409a3aa55279fab38b3dfbfcdeb1a96
https://github.com/shagu/pfUI/commit/6c030db7208b9d463cdd463baefb5bde95f1866a

Function is called: `pfUI.uf:ClickAction(button)` where `this.label` and `this.id` give informations about unitframe type.

kxseven:
Awesome!

In order for Clique and other addons like it to work the `pfUI.uf:ClickAction(button)` function needs to check first for a valid custom function being hooked and then call it if needed before falling back to the 'default' click code/behaviour

Adding the below at the top of the  `pfUI.uf:ClickAction(button)` should do the trick

```
-- use the custom OnClick function if one is hooked
if (type(pfUI_Custom_ClickFunction) == "function") then
    if (pfUI_Custom_ClickFunction(button, this.id)) then
        -- pfUI_Custom_ClickFunction should return true if handled, then we do nothing more.
        -- no return and we'll continue with default behaviour
        return true
    end
end
```

shagu:
Do you want me to add this? because I won't do it. I might got you wrong, but every addon that hooks functions have to hook it on their own. I will not provide a pfUI_Custom_ClickFunction.

Use this in any other addon code to hook the ClickAction:
```
if pfUI.uf.ClickAction and not pfHookClickAction then 
  pfHookClickAction = pfUI.uf.ClickAction

  function pfUI.uf:ClickAction(button)
    -- your code here
    --[[ foo bar ]]--

    -- run the original function
    pfHookClickAction(button)
  end
end
```

PR #436, opened 2017-02-07
Added support for custom OnClick hooking

kxseven:
- Adds a dummy ping() function that can be used to check if pfUI exists and is loaded
 - Checks for existence of pfUI_Custom_ClickFunction and calls it if hooked before fallback to default pfUI OnClick behaviour (which enables Clique, HealBot, etc to work)
 - Clique with pfUI support required (https://github.com/kxseven/Clique)


shagu:
1. I will never merge dirty PR's like this. If you have something you want to see in pfUI, rebase your work to the master branch and create single PR's for every feature.

2. I already told you here (https://github.com/shagu/pfUI/pull/419) that I won't add your hooks inside of pfUI. The function is as it is right now perfectly hookable for every other addon. I even gave you an example on how to do it.

3. Why do you need a ping function? You could just check for "pfUI" instead or any of the hundreds of other methods. Also, relying on a global variable with the name "ping" is in general a bad idea.

4. Your mobhealth nameplate code won't work, because your checking the target health, but not all of your nameplates are your actual target.

Nevertheless, thanks for your effort and please keep up your motivation ;)

PR #442, opened 2017-02-08
GUI Optimization #1

Cruzix:
I've slightly edited how the features in the pfUI settings are explained so that it's easier to understand what they do.

I also noticed that a lot of options can be toggled in more than one section, so I'll be working on a "GUI Optimization#2" to remove all the duplicates and to relocate the options from the Appearance section so it can be removed since there's no need for it.


shagu:
Awesome! Thanks mate

shagu:
https://github.com/shagu/pfUI/commit/c7d565b3af5b9101a738e2119391e97e546230fc

PR #520, opened 2017-05-17
Fixed bug with player coordinates inconsistency

greenya:
There is a small bug with player coordinates inconsistency:
1) mouseover minimap - the minimap corner shows one coords;
2) mouseover location title under minimap - the tooltip shows other coords;
Please see screenshot -- http://i.imgur.com/RdboPjW.jpg

This is due to different rounding algorithms.
So i made this PR wich actually does next:
1) rounding now the same -- via "round";
2) formatting now the same -- comma separated (it was "/" for "panel");
3) "minimap" has kind of error checking -- "posX ~= 0 and posY ~= 0", so i added this checking to "panel" too;

### DIFF
--- a/modules/minimap.lua
+++ b/modules/minimap.lua
@@ -115,9 +115,7 @@ pfUI:RegisterModule("minimap", function ()
     SetMapToCurrentZone()
     local posX, posY = GetPlayerMapPosition("player")
     if posX ~= 0 and posY ~= 0 then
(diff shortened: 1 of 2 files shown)


shagu:
https://github.com/shagu/pfUI/commit/38581f724db7901741ac72a9b2b11eede6b28b5e

Thank you =)

Issue #365, opened 2017-01-18
chat: option to show channel names


CoolFangs:
Unless I'm missing it somehow, there doesn't seem to be any option to enable channel names in the chat.

Rosexephon:
Would also like to see this happen, as well as be able to put different channels in different windows.

PR #534, opened 2017-05-22
chat: Enable Channel Numbers Only option

greenya:
Related to issue #365

This PR adds option to enable channel numbers only mode (which is right now the only mode). When UNCHECKED, channel names will be included with their numbers. Please see screenshot -- http://i.imgur.com/ZZeSjwH.jpg


shagu:
https://github.com/shagu/pfUI/commit/83bb05f6160b221d0c4d15c00057baf382b1130d

PR #568, opened 2017-06-22
Added localized files to make it easier for Chinese people to use!

aixiu:
In order to facilitate the use of the Chinese people, we luacn.net community for the pfUI increase the localization of the document, because too many people like, so hope shagu can be incorporated into the localization file, thank you.


shagu:
Hi, Thank you very much for the effort you put into this. 
I've introduced a new API which handles translation in a more convinient way to create translations as easy as possible. 

It would be really great if you could rebase your work and insert your translation in here:
https://github.com/shagu/pfUI/blob/master/env/translations_zhCN.lua

aixiu:
It's really great to do that. Our small team has been changing the source variables for a long time. Your introduction of the new API interface this time has made our team much less productive. Thank you

aixiu:
OK, when I've finished localization, submit the PR

PR #573, opened 2017-06-26
Fps panel

Sqoops:
1. Fix double fps calculation by just using the local fps variable.

2. Added addons-memory calculation to the fps counter on the pfui panel. This shows how much MB ram your addons currently consume.

### DIFF
--- a/modules/panel.lua
+++ b/modules/panel.lua
@@ -168,7 +168,9 @@ pfUI:RegisterModule("panel", function ()
 
       local _, _, lag = GetNetStats()
       local fps = floor(GetFramerate())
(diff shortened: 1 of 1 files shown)


shagu:
Thanks for your input and your preliminary work on this. But I don't really like too many values within one field. That's why I did it slightly different:
- The memory usage is now shown in the tooltip
- The next garbadge collector run is also shown in the tooltip
- using "round" instead of "floor" to provide 2 decimal places

https://github.com/shagu/pfUI/commit/bf2a6c06e4181ab3ddc435a1a7a63dd3c993217e

PR #580, opened 2017-07-04
unitframes: health & power texts

Linae-Kronos:
This allows to have 3 texts on Health Bar & Power Bar, so 6 texts in total. Works for every unitframe : player, target, target of target, pet, focus, group, group target, group pet, raid.

pfui_unitframes_preview


Grookshank:
That would be fantastic if merged.

PR #597, opened 2017-07-18
Fix localization errors and add a localization

aixiu:
Fix localization errors and add a localization


shagu:
thanks :)

https://github.com/shagu/pfUI/commit/c87db7a712adf79b5fa58cf40ccdc0251ffb6710

PR #599, opened 2017-07-18
SetFrameStrata

Sqoops:
Set Frame in BACKGROUND to not conflict with other addons.


shagu:
https://github.com/shagu/pfUI/commit/c7da2aa8d074606c6c1698e3de837ae2b1787ba9

Issue #617, opened 2017-08-07
BUG - undocked chat windows reset on UI reload


Priddles:
### Issue Severity - LOW
This issue:
- does not impact many users
- has a simple workaround
- is mainly an inconvenience; no other features appear to be affected

### Expected Behaviour
Undocked chat windows will retain their new positions and dimensions.

### Actual Behaviour
Undocked chat windows will retain their new positions and dimensions only until the next UI reload.

Whenever the UI is reloaded:
- the positions and dimensions of all undocked chat windows are set to the "General" chat window defaults
- all chat windows correctly retain their "lock" state

Probably unrelated but noticeable behaviour is that:
- all chat windows are rendered in the order of their creation, such that the last created chat window is "on top" of the rest
- all chat window interaction is layered in reverse order of their creation, such that the last created chat window is "underneath" the rest

This can be confusing when the user tries to move their custom chat window (which looks like it's on the top) but ends up moving the "General" chat window instead (assuming its unlocked).

### Workaround
Whenever the UI is reloaded, reposition and resize the chat windows. This workaround is time consuming and can be made more difficult due to the "layering" of the chat windows and the fact that they retain their lock state.

### Tested Configurations
**Config 1:**
- Client - US v1.12.1
- AddOns - whatever I had the first time I tried to have an undocked chat window
- WDB - same as AddOns
- WTF - same as AddOns

**Config 2:**
- Client - US v1.12.1
- AddOns - pfUI (commit e749b9543039f6bf96028b5716fb39199d117fed)
- WDB - empty
- WTF - empty

### Comments
Yes, this was completely unnecessary ^.^

Road-block:
Holy Smokes!  
Overkill, but A++ for effort.  
This would make an excellent template for the *perfect* bug report 🥇 🙇

Priddles:
### The Problem
After reading through the code, I think I know what's happening.

The cause of the bug is in `modules/chat.lua::RefreshChat`, more specifically `lines 315 ~ 328`:
``` Lua
if i == 3 and C.chat.right.enable == "1" then
  tab:SetParent(pfUI.chat.right.panelTop)
  frame:SetParent(pfUI.chat.right)
  frame:ClearAllPoints()
  frame:SetPoint("TOPLEFT", pfUI.chat.right ,"TOPLEFT", default_border, -panelheight)
  frame:SetPoint("BOTTOMRIGHT", pfUI.chat.right ,"BOTTOMRIGHT", -default_border, panelheight)
[...]

PR #618, opened 2017-08-08
Close issue #617 and add "Hide Combat Log" feature

Priddles:
Implemented trivial solution for issue #617 and added a "Hide Combat Log" feature.

`RefreshChat()` now only resets the position and dimensions of the "General" and "Loot and Spam" chat frames (ChatFrame1 and ChatFrame3 respectively).

Added a checkbox in the GUI named "Hide Combat Log" that, when checked, will `UnDock` and `Close` ChatFrame2 on UI reload.


Priddles:
@shagu Please note that you may want to change my implementation of the "Hide Combat Log" feature.

I was trying to keep changes to a minimum to prevent merge conflicts, but I feel that it has negatively impacted the code. Personally, I think the logic should be moved into `RefreshChat()` instead of polluting `SetScript("OnEvent", ...)`.

You may also want to refactor the config entry (currently `C.chat.combatlog.hide`). I wasn't sure which of the current subgroups it would belong to (`global` maybe?) and just decided to make a new one instead.

shagu:
Thanks https://github.com/shagu/pfUI/commit/c4996bc35bee1633c547ac342d69e3cb5d04cae4

shagu:
https://github.com/shagu/pfUI/commit/77d357ea06d4a429a008f40396e8f1c88a6ebb44

PR #628, opened 2017-08-19
Updating ruRU localization

Lichery:
I apologize if something has not done so. The first time I do pull request


Lichery:
It seems to have turned out. xD

PR #670, opened 2017-09-24
Update panel.lua

unknauwn:
ignore hunter ammo bag, soul bag & special bag like herbalism, ench etc
Show only Regular bag "free slot (usedslots/maxslots)"


shagu:
Thanks sir, but I made it like that:
https://github.com/shagu/pfUI/commit/c01ff76e7c592622332a3709872f30256081083a

=)

PR #671, opened 2017-09-24
Adding Memory Addon Usage in description

unknauwn:
Adding Memory Addon Used in tooltip panel fps/ms


shagu:
It's already there:
https://github.com/shagu/pfUI/blob/master/modules/panel.lua#L159

Issue #666, opened 2017-09-23
skin: add support for underwater breath bars


nesherman:
Best to see the attached image: https://imgur.com/a/N1a14

Thanks, and have a great day!

lbjdaking23:
Official name for these bars seem to be MirrorTimers. 
A [link](https://www.townlong-yak.com/framexml/1.12.1/MirrorTimer.xml) to the XML design for them. 
I am going to see if I am going to try my hand on skinning them unless @shagu wants to.

shagu:
Thank you for your comment, @lbjdaking23. Feel free to have fun with those frames. Pull Requests are always welcome.

shagu:
Thank you very much for your contribution @lbjdaking23 !

PR #680, opened 2017-09-26
skin: add mirror timers

lbjdaking23:
Preview: 
oakkftgc

This is can definitely be improved in the future like making the frames movers and being able to customize them.

Fixes issue: #666

### DIFF
--- a/modules/mirrortimers.lua
+++ b/modules/mirrortimers.lua
@@ -1,0 +1,12 @@
+pfUI:RegisterModule("mirrortimers", function ()
+  for i = 1, MIRRORTIMER_NUMTIMERS do
+    _G["MirrorTimer" .. i]:GetRegions():Hide()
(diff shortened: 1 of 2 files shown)


shagu:
Thank you, they look really nice! May I suggest putting this into the skin module instead of a separate one?

shagu:
Thank you, they look really nice! Did you plan to work on adding the mentioned improvements? Would be also nice to have timer on them and I'd like to suggest a slightly bigger padding between border and the text.

*PS: Within pfUI modules, you don't have to type pfUI.api infront of CreateBackdrop*

lbjdaking23:
I can definitely add the time to them. Also will look into increasing the padding.
For other improvements I will see if I can figure them out not too sure because I am a lua/wow api as you can see with the CreateBackdrop call 😛

PR #682, opened 2017-09-27
Loot module additions (masterloot)

Road-block:
It's not particularly clean I'm afraid (hate default UIDropDownMenus with a passion)


api/api.lua
```diff
@@ -183,9 +183,73 @@ function pfUI.api.Abbreviate(number)
   return number
 end
 
+-- [ Widest Audience ]
```
shagu: I'd like to suggest a function name that is something more meaningful and more related to existing game functions. Maybe something that is more like `SendChatMessageWidest`, `SendChatMessageRaid` or whatever.

api/api.lua
```diff
@@ -183,9 +183,73 @@ function pfUI.api.Abbreviate(number)
...
+-- Sends a message to highest channel the player can broadcast to
+-- 'msg'        [string]          the message to send
+function pfUI.api.widestAudience(msg)
+  local channel = "SAY"
+  if UnitInRaid("player") then 
+  -- Roadblock: maybe RW too spammy, try with just RAID
```
shagu: This comments shouldn't go into the master branch.

api/api.lua
```diff
@@ -183,9 +183,73 @@ function pfUI.api.Abbreviate(number)
...
+        end
+      end
+    end
+  end  
+end
+
 -- [ hooksecurefunc ]
 -- Hooks a global function and injects custom code
--- 'name'       [string]           name of the function that shold be hooked
+-- 'name'       [string]           name of the function that should be hooked
```
shagu: Ooops, thanks :D

api/api.lua
```diff
@@ -211,6 +275,43 @@ function pfUI.api.hooksecurefunc(name, func, append)
   _G[name] = pfUI.hooks[tostring(func)]["function"]
 end
 
+-- [ EnQueue ]
```
shagu: I'd like to suggest another more meaningful function name, maybe QueueFunction.

modules/gui.lua
```diff
@@ -1113,6 +1113,7 @@ pfUI:RegisterModule("gui", function ()
       CreateConfig(this, T["Enable Auto-Resize Loot Frame"], C.loot, "autoresize", "checkbox")
       CreateConfig(this, T["Disable Loot Confirmation Dialog (Without Group)"], C.loot, "autopickup", "checkbox")
       CreateConfig(this, T["Enable Loot Window On MouseCursor"], C.loot, "mousecursor", "checkbox")
+      CreateConfig(this, T["Enable \'by Class\' MasterLoot Menu (and extras)"], C.loot, "classmasterloot", "checkbox")
```
shagu: If you don't mind, I'd rather call that "Enable Advanced Master Loot Menu".

modules/loot.lua
```diff
@@ -17,6 +17,375 @@ pfUI:RegisterModule("loot", function ()
     UpdateMovable(pfUI.loot)
   end
 
+  pfUI.loot.rollCapture = string.gsub(string.gsub(string.gsub(RANDOM_ROLL_RESULT, "[()]", "."), "%%s", "([^ ]+)") , "%%d", "(%%d+)")
```
shagu: I didn't relly checked this one, but would https://github.com/shagu/pfUI/blob/master/api/api.lua#L88 have the same effect?

modules/loot.lua
```diff
@@ -17,6 +17,375 @@ pfUI:RegisterModule("loot", function ()
...
+        if i == to or k == 15 then 
+          EnQueue(widestAudience,names)
+        end
+        k = k<15 and k+1 or 1
+      end
+      pfUI.loot:RegisterEvent("CHAT_MSG_SYSTEM")
+      pfUI.loot.randomRolling = true
+      EnQueue(RandomRoll,"1",tostring(to))
+    end
+  end
```
shagu: I'd love to see an empty line between function definitions

api/api.lua
Road-block: absolutely, I wasn't expecting a straight up merge, just wanted to get something functional up for comments (like this) and some decision as well. 
Do you want it to go to RW when the broadcaster has the rights or is that too "loud"?

api/api.lua
Road-block: yep.

modules/loot.lua
Road-block: quite possible, I started working on this before that specific commit.

Road-block:
Alright so I don't reply to every thing separately :)  
I'll try to clean it up a bit and implement your suggestions in the process, but at the end I don't think I'll be able to get it to your standards exactly.  
So feel free after I made whatever changes to it if it's still not satisfactory to do any edits you want or even just pick what you want and integrate it, I don't care that I show up on the history.

Road-block:
What this thing does in a nutshell.

- Class Grouping
- Extra Utility

The RL can set banker, disenchanter from the unitframes context menu and then find them in the Special Recipient group for assigning items directly.  
Random does what you'd expect: prints out the names and numbers of the raid members and then does a /roll and assigns to winner.  
Request rolls, starts monitoring for rolls from the raid and shows them sorted high > low (reopening that group needed)

modules/loot.lua
Road-block: It is not exactly the same the current version produces this pattern  
`([^ ]+) rolls (%d+) .(%d+)-(%d+).` your function produces this  
`(.+) rolls (%d+) %((%d+)-(%d+)%)`  
But in practical terms the captured bits will be the same (since no player names will ever have spaces in them) so I'll use your function.

shagu:
Thank you very much! And also thanks for the cleanup ;)

I only made a few other minor changes on that:
* removal of trailing spaces
* renamed config entry to a shorter string
* enabled by default

-> https://github.com/shagu/pfUI/commit/a2e0615659c4d4db5b291a0a4377e7de1652673e

PR #693, opened 2017-09-28
minimap: improve zone text options

lbjdaking23:
This change easy allows the zone text panel to be hidden which was requested on the [forums](https://forum.elysium-project.org/topic/29034-addon-pfui-an-elvui-inspired-addon-entirely-written-from-scratch/?page=15&tab=comments#comment-442937).


lbjdaking23:
I just been told by someone their is a option to change the text that shows up in that location. I can fixes this in a better way imo then having this more confusing change.

PR #702, opened 2017-10-01
Masterloot. Fallback for GetMasterLootCandidate not updating immediately for leavers.

Road-block:


Road-block:
Just a note that the other option would be to remove or disable the button for the unit which does not have unit information (is no longer part of the group but the client hasn't yet removed it from the MasterLootCandidate index).  
The reason I did not choose to go that route is not from a coding perspective, removing invalid masterloot targets would obviously improve the game (see GetMasterLootCandidate [notes](https://wow.gamepedia.com/API_GetMasterLootCandidate) at the bottom of the page) ; it is for group dynamic - social reasons.  
In that it is preferable to have an invalid masterloot target and have the masterlooter and group realize that the player has left the group than have them question whether valid masterloot targets are erroneously removed.  
In other words the fix is siding with the suspicious mind 😛 

Also to note that the masterloot index is refreshed every time the dropdown is shown, there is no caching of indexes (if the wowwiki note would worry you in that respect).  
The masterloot index of the addon is **always** current as far as the client is concerned.

Road-block:
Adding a blue post from back in the day that shows the quirkiness of GetMasterLootCandidate.  
http://blue.cardplace.com/cache/wow-interface-customization/327106.htm

Issue #656, opened 2017-09-19
actionbar: add classic snowfall alike behaviour


nesherman:
An option to use abilities based on keystate DOWN rather than keystate UP would be a very welcome addition to pFUI. Currently many of us are using Snowfall to achieve this functionality. Thanks, and have a great day.

shagu:
https://github.com/shagu/pfUI/commit/6f7f8a18b485e1604921e1f3657d08c9a05214e6

PR #711, opened 2017-10-14
Actionbar: Action on button down (toggle)

Sour:
Can enable/disable in settings.
#656


shagu:
I'm sorry of the delay. I am aware of this PR but like to have some detailled testing on that. Right now, I'm very busy with all the other addons and holding this back. Please be patient, Thank you :)

rhnorskov:
I have been doing some testing, and it seems to working great, with the exception of `Auto Attack` and `Auto Shot`. If a keybind for any of the two abilities is pressed an `Interrupted` error will be thrown. Maybe an exception needs to be added to these two abilities.

shagu:
This should do the same, or am I missing something?
```
  if C.bars.keydown == "1" then
    hooksecurefunc("ActionButtonDown", function(id)
      ActionButtonUp(id)
    end, true)
  end
```

shagu:
It also seems like, your solution does only work for the main actionbar, I've added another hook for the MultiActionBar:

```
  if C.bars.keydown == "1" then
    hooksecurefunc("ActionButtonDown", function(id)
      ActionButtonUp(id)
    end, true)

    hooksecurefunc("MultiActionButtonDown", function(bar, id)
      MultiActionButtonUp(bar, id)
    end, true)
  end
```

shagu:
https://github.com/shagu/pfUI/commit/6f7f8a18b485e1604921e1f3657d08c9a05214e6

Issue #720, opened 2017-10-22
group/raid frames: missing indicator for buffs


Grookshank:
pfUI displays some buffs your class can cast on group frames, but not all it seems. I am missing:
* soulstone
* fire shield
* blood pact

This is the icon indicator I mean.
https://imgur.com/tcSXBsF

shagu:
https://github.com/shagu/pfUI/commit/9708ac50d1582c8a09eedba9b95ffd235c892ed9

PR #721, opened 2017-10-22
Fix warlock buff missing in group frames

Grookshank:
Fix warlock buff missing in group frames.

https://github.com/shagu/pfUI/issues/720


shagu:
Fixed trailing spaces, empty lines and merged here: https://github.com/shagu/pfUI/commit/9708ac50d1582c8a09eedba9b95ffd235c892ed9

Thanks

Issue #253, opened 2016-12-31
Snap to grid for moving elements


Cephel:
If you're like me, you can't deal with elements being like a quarter of a pixel off of being "perfect". A configurabe gridsize complete with snapping elements to intervals of this size would help towards this.

PR #727, opened 2017-10-26
unlock: Snapping to other frames

greenya:
snapping

### This PR is about following
1. Removed redundant loop **for i = 0, 128 do** as its body has nothing to do with the index, so that was probably a mistype/copy-paste; the body left untouched, and now it executes only once (please see the diff, line 35)
2. Replaced **pfUI.unlock.settingChanged = true** with **pfUI.gui.settingChanged = true** as I also think this is mistype, since couldn't find any _settingChanged_ with _unlock_, only with _gui_ (please see the diff, line 286)
3. Added snapping to other frames when moving a frame; this feature enabled only when user start moving frame with ALT button pressed

Please note, i haven't changed the help text (see pfUI.unlock:UnlockFrames), because its translated on many languages, so i kept it untouched; would be good to add something like "Hold Alt to enabled snapping to other frames" or similar.

P.S.: this PR is partially related to issue #253

### DIFF
--- a/modules/unlock.lua
+++ b/modules/unlock.lua
@@ -32,12 +32,10 @@ pfUI:RegisterModule("unlock", function ()
 
       local height = GetScreenHeight()
 
-      for i = 0, 128 do
-        local tx = pfUI.unlock:CreateTexture(nil, 'BACKGROUND')
-        tx:SetTexture(.1, .5, .4)
-        tx:SetPoint("TOPLEFT", pfUI.unlock, "TOPLEFT", 0, -(height/2) + (size/2))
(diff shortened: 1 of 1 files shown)


shagu:
Thank your for this PR. But unfortunately it doesn't seem work correctly to me. The frames do always overlap each other whenever they're docked. I guess you'll have to take care of the default_bordersize or even better, using the sizes of frame.backdrop if existing. (https://github.com/shagu/pfUI/blob/master/api/api.lua#L585)

greenya:
Tested it more; you are right, it has flaws with different border sizes.
I'll let you know if anything viable comes up.

PR #731, opened 2017-10-28
Make buff/debuff size in group frames configurable

Grookshank:


shagu:
I have kept you as author, but have renamed the options and merged it here:
https://github.com/shagu/pfUI/commit/3aec8e2f58b90d6cfe274f423562e4a08d5f2016

Thanks :)

PR #733, opened 2017-10-28
Fixed stratas that was set wrong. 

stokbaek:
Some UI elements overlapped some places were they shouldn't. This should fix the problem.


shagu:
I'm very careful with frame-strata fixes nowadays, because the latest changes had always caused different addons to misbehave and led to a revert. Are you sure this won't break anything?

stokbaek:
I tested it and running with it myself. It is only some minor UI elements I changed so the dont do some wierd overlaps

shagu:
docked SWStats and KTM will overlap the panel and behave weird on bags:
bags

panel

Issue #699, opened 2017-09-30
minimap: skin tracking symbol


Grookshank:
Would look better if they are:
- sized the same
- square like everything else in pfUI
- have a black/grey border like everything else
- positioned on the edges of the squared pfUI minimap instead of a cirlce

Also nice would be to be able to move them on the squared minimap border.

shagu:
obsolete.

Issue #208, opened 2016-12-23
Enhancement: Add drop down menu feature for minimap


Cruzix:
Add an drop down menu for the minimap when you right click (or middle mouse) on the minimap. 

This drop down menu should contain stuff such as hunter/mining/herb tracking abilities and maybe even profession short cuts. This is one feature that I miss having from current WoW and hopefully you can make it happen.

nesherman:
Also second this.

phendrome:
Mhmm... I feel the same way here. The mail button is nice, when you have new mail. Blends in nicely.

The tracking button for mining, herbalism, whatever, is quite off. Possible to fix?

PR #766, opened 2017-11-22
tracking: module added

greenya:
Hello.
This PR adds new module -- **tracking**, it replaces default Minimap Tracking Icon.

t1

### Basic features (just like default Tracking Icon does):
- if nothing is tracked, icon disappers;
- if mouse over, correct tooltip shown;
- if right mouse click, you cancel tracking;

### Additional features
`pfUI Settings -> Settings -> Apperance`
t2
- you can change border size of the frame (default is -1, meaning to use Global Border Size);

`pfUI Settings -> Minimap -> Minimap`
t3
- you can change size of the icon (default is 20);
- you can enable indication of "no active tracking" state (useful when you keep forgetting to use tracking after character died); this is not blizz-like feature, so its disabled by default; if enabled, the icon never disappear, become a blinking question mark in case no active tracking; please note: if you dont have any tracking abilities in your spell book, turn this feature off, because right now it doesn't check it, resulting in never ending indication;
- you can move tracking icon frame using **Unlock** module; by default the frame is locked to minimap, so if you move minimap, it will move with it (just like "minimap panel" does); if you move icon separately, you unlock it from minimap (just like "minimap panel" does);

### Future plans for extending the module
- ability to check spell book for "tracking" spells (so no blinking if you can't track anything);
- ability to left mouse click, see list of available "tracking" spells, clicking on any of them should "/cast" the spell;

### Related issues
- #699, current implementation fully covers it;
- #208, not covered, only in "future plans";


greenya:
Implemented drop down menu when player left mouse click the icon
tdd

shagu:
Thanks alot, I made some changes to it and merged it with this commit: https://github.com/shagu/pfUI/commit/50dd24ec4b43be4aa00966975b3189e92990cfdc

PR #778, opened 2017-12-02
Unitframes: Option for vertical/horizontal HP bars

stokbaek:
I have tested player and raid. They work with vertical and horizontal with and without invert + healcomm.


shagu:
I'd suggest to:
* change the config entry from "healthbar_dir" to "verticalbar" or something like that
* make the available options be "0" and "1" 
* set it to type "checkbox" in the GUI
* changing the Caption from "Horizontal/Vertical Health Bar" to "Enable Vertical Health Bar"

As there is only horizontal and vertical possible, I doubt that someone will add "diagonal bars" at any point 😁 

Very nice PR btw, thanks for that.

stokbaek:
I was in conflict if I should do it like I did or if I should use checkbox, if you find it better as checkbox will I change it tomorrow :)

PR #788, opened 2017-12-15
CombatFeedback on unitframes

Jayrgo:
added combat feedback to unitframes (player, pet, target, group, raid)


shagu:
Thanks for this suggestion, I was playing around with this trying make the combat-message selectable, like all the other unitstrings like "Health - Min/Max" and so on. Might take a while.

Jayrgo:
you can hook "CombatFeedback_OnCombatEvent" to filter the message or adjust the colors

shagu:
Thank you. I have splitted and rebased your commit to match the latest API.
It's available on `master` with this commit: https://github.com/shagu/pfUI/commit/50eff475d9cd918f097beff641111e1616f4ef84

PR #795, opened 2017-12-18
loot: Fix Set Banker and Disenchanter UnitPopup entries.

### LINKED ISSUE #791: loot: can't set banker and disenchanter if target not in group
See title: When I try to set a banker and disenchanter, the options only show when the player is in my group. They do not show, if they are not in my group, but in the same raid.

Road-block:
loot: Inject the banker and disenchanter entries to another couple relevant UnitPopup menus.  
Fix the unitpopup conditional hide so it actually works and those options do not display when not relevant.
Closes #791


shagu:
@Road-block is this safe?

PR #815, opened 2017-12-29
updated esES

woopydalan:


stokbaek:
You dont have to make a new PR every time you make a commit. It should automatic be added to the PR when you commit it I think

env/translations_esES.lua
```diff
@@ -15,11 +15,11 @@ pfUI_translation["esES"] = {
   ["Always Use 2D Portraits"] = "Siempre Usar Retrato 2D",
   ["Ammo Counter"] = "Contador de Munición",
   ["Appearance"] = "Apariencia",
-  ["AtlasLoot"] = nil,
+  ["AtlasLoot"] = "AtlasLoot",
```
shagu: You don't have to translate if the translation has the same value. Just keep it as "nil"

shagu:
https://github.com/shagu/pfUI/commit/8c17a48271acd746a51bfdbd0130b7c28adfe3be

PR #819, opened 2017-12-31
Announce players ress in chat

stokbaek:
If you set option to "Raid" and you are not in a raid will it yield a warning/info text about not being in raid, maybe a check needs to be added that checks if player is in a raid before message is being sent.

Tests:
- [X] Say
- [X] Party
- [X] Raid
- [X] Dynamic
- [X] Battleground


Road-block:
There is a method already for announcing to highest available channel if that's of any use.
[api/api.lua:204](https://github.com/shagu/pfUI/blob/f2543d71af01d995a78e9b2d19ba985f34e000f7/api/api.lua#L204)

stokbaek:
Thanks for finding that function, but I dont think it would be useful in this case. It will not cut down the lines of code used, and it might just confuse others when two different chat methods are used.

I have tested in BGs and it should also work here also.

Road-block:
I found it because I wrote it. 😛   
What does "cut down the lines of code used" mean? Lines of code are not a useful metric of anything, except when you have similar code in several places that could be factored out to a generic method (which is what is generally in `api`)  
I don't think your code will work like you think anyway.  
Personal opinion.  
Use the existing method for the smart announce and only give 2 options:
- Smart (will announce to party or raid or say depending on availability)
- Disabled (no announce)

stokbaek:
But what if a person wants it to be said in say or party when in raid and so on.

I see plenty of people using say while in raid to not fill raid chat because of loot council talk for example.

Road-block:
Your code becomes
```lua
if C.unitframes.ressannounce ~= "disabled" and sender ==  UnitName("player") then
  SendChatMessageWide("Resurrecting " .. target)
end
```
and your config a simple toggle (checkbox)

stokbaek:
closing this PR as I have made it as its own addon https://github.com/stokbaek/ResurrectionAnnounce

PR #825, opened 2018-01-04
bubbles: add speaker and custom callout texture

Road-block:
Graphic work is an abomination and it's above my abilities to improve it.  
But I wanted to share idea for small improvements to the new chat bubble module.  
I like the speaker feature and I think that part is fine as is, I would like a custom callout too that replaces the default chat bubble tail that looks roughly like a straight line towards the speaker but I cannot make it look better 😛 
image

Basically treat this as a prototype.


Grookshank:
I think it looks not so good, since the line is straight from the center or the bubble. Diagonal would looke better.

Personally I think though, it is more clutter than worth it; same with the names.

Road-block:
How it looks with the current commit
image

Grookshank:
Looks better imho. I would suggest to try thinner diagonal line.

I am still not sold on the names. How does that interact with the players names option?

Road-block:
> I am still not sold on the names. How does that interact with the players names option?

It doesn't. Player / NPC names are usually covered up by chat bubbles (same happens on default client)  
The second reason why one *might* want a speaker on the bubble (as an option) is that when there are 2-3 chat bubbles close together its oftentimes hard to know which bubble belongs to who without looking at the chatframe at the same time (which kinda defeats the purpose of having chat bubbles showing in the first place)  

In any case like I said in the first post, treat this as a prototype, I'm not suggesting my graphic work as anything worthwhile (I suck at it).  
If someone wants to contribute a better callout texture feel free to chip in.

shagu:
image
image

Grookshank:
Very nice.

shagu:
Still not happy with it.. Not sure if I'm going to include it.
image

Road-block:
image

Road-block:
You can close this, just give me that bubble callout that you don't like - *that I like 😋 * - so I can have it on my local copy 💯

PR #840, opened 2018-01-19
unitframes.player: Add Weapon Buff indicators anchored LEFT. (prototype)

Road-block:


shagu:
i like it! Thanks for sharing
* would be great to have an option to include the weaponbuffs in the other buffs
* would be greate to have a seperate pull request for the `pfUI_config` -> `C` fixes 😁

PR #876, opened 2018-04-06
Fix to horizonfarclip console command so that it adjusts to the upper limit of 2112.

EsWoW:
`HorizonFarClip must be in range 1305.000000 - 2112.000000.`


shagu:
Hi, may I ask where this quote is from?

EsWoW:
This is from the console when loading the WoW.exe with the -console flag and then opening it up with the grave key (or whatever you rebound it to). If you enter in a range outside of what is accepted, it lets you know.

PR #891, opened 2018-04-21
Added option to show totems as procs on frames

stokbaek:
This allows shamans to set totems to show as buff indicators on the frames.


shagu:
Looks good, but please remove the check inside the gui module. It's fine to have that shown for all classes.

PR #903, opened 2018-05-02
fixed durability calculation of items

Artur91425:
the calculation should be the arithmetic mean of the interest amount.
(before it always wrote the lowest percentage of item)

also added small code optimization (removed unnecessary characters)

### DIFF
--- a/modules/panel.lua
+++ b/modules/panel.lua
@@ -1,4 +1,4 @@
-pfUI:RegisterModule("panel", function ()
+pfUI:RegisterModule("panel", function()
 
   local font = C.panel.use_unitfonts == "1" and pfUI.font_unit or pfUI.font_default
   local font_size = C.panel.use_unitfonts == "1" and C.global.font_unit_size or C.global.font_size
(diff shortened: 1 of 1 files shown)


Road-block:
I disagree on this.  
Every durability addon in existence shows the lowest item and the reason is mean is pointless.  
You want to know when you lost a slot, cause a red item is as if you unequipped it and have nothing in that slot.  
A mean or whatever could be a secondary value shown but lowest item is absolutely the right info to show.

Lichery:
we in fact consider the general item durability, and not any specific, and when item durability is small, then this item will be hightlighting on DurabilityFrame. I still think that you need to consider the general to know the average state of the equipment, rather than a specific item.

Road-block:
I'm not saying that some *extra* information would go amiss, but you **need** to know that your engineering shield is broken, or your weapon melted on Ragnaros or x item is about to break.  
I'm just disagreeing that mean durability should be the primary / only information.  
The general disrepair of your gear is also shown as the mouseover repair amount.  
DurabilityStatus as well as the various Titan and FuBar durability plugins all show lowest item in the main display and a per-item breakdown or average in the tooltip / extra details.  
I believe this (the need to know at a glance that some item is at 10% and next death will make it useless) is the reason.

Do I really care that the average durability of my gear is 60% if I have 2 items at 10%?

Lichery:
If this is interesting, I can share code. I can not finish. Now there is a problem with updating the tooltip. It does not update to onEnter :( The update only occurs after the reload UI

Lichery:
Also can play around with colors and adjust the visualization of the percent display

shagu:
Hi @Lichery I appreciate what you did. But I'm with @Road-block at this point. It's way more useful to see the lowest durability because this gives the required info when something is about to break.

If we find more people who would like to see the average-durability, like you do. Then we can make an option for that, for sure. But otherwise, I'd just close this one.

**Anyhow, The tooltip screenshot you've posted looks awesome! I would love to see a PR for that.**

PR #904, opened 2018-05-03
Added tooltip for durability

Lichery:
I remind you that this has not been completed yet. Update does not work on onEnter.  Update works only after reloadUI!

it is also possible to set other ranges and colors for percentages of durability


Lichery:
Sorry, I realized that I did wrong ... I will redo it now and it will work as it should.

Lichery:
Did, it works perfectly. But once again I apologize, I will close this PR and re-create it to be neatly, because I rewrited a lot of code.

PR #905, opened 2018-05-04
Added tooltip for durability

Lichery:


Lichery:
The only problem remained. For some reason, when you first enter the world (after close game client) there is no calculation of lowestPercent. More precisely, scanning does not begin ... as if the problem is in event, but I looked and function UpdateRepair() is called on PLAYER_ENTERING_WORLD

Lichery:
By the way this bug is in the current version

Lichery:
Please help to solve the above problem. Outputting debug messages I was able to find out what the problem is in repairToolTip.

1. Problems begin in the line `hasItem, _, repCost = repairToolTip:SetInventoryItem("player", id)`. When you first enter in the world, hasItem returns the correct data, but repCost returns 0 for all items and after the reloadUI starts to return the correct data.
2. Before the cycle `for i = 1, repairToolTip:NumLines()`, it works correctly, but inside the cycle it stops working.
3. If we replace `repairToolTip:NumLines()` by a number (e.g. 10), then inside the cycle begins to work, but again stops working on the next line `tmpText = _G["repairToolTipTextLeft"..i]`

Share this mini add-on for testing. Please see if you can understand what the reason for this behavior is and correct it.
[test.zip](https://github.com/shagu/pfUI/files/1976709/test.zip)

Lichery:
wow 2018-05-07 01-37-36-950

Lichery:
I found a problem with the channel spell. The last used texture is set on them. I tried to disable the icon and override the points, but I can not do it so that everything will work properly to work. In the end, I was able to make such a patch... https://github.com/shagu/pfUI/pull/905/commits/b50ba42708893bf026f7105f5f90eff737515e99

Lichery:
wow 2018-05-07 19-43-08-535

Lichery:
I'm sorry that everything is in one place, but I can not create a new PR, if there is already one

Lichery:
I can not do the right PR... Can you transfer this commits?
https://github.com/Lichery/pfUI/commit/6e99cf1c34313b5a8c09706cde651e9786a26629
https://github.com/Lichery/pfUI/commit/a0cc14a8c66315b89b129714f245b83ea6681dda

shagu:
Thanks, I've chosen @Road-block 's implementation of the panel tooltips. I've added this one aswell:
https://github.com/Lichery/pfUI/commit/6e99cf1c34313b5a8c09706cde651e9786a26629

PR #907, opened 2018-05-05
Implement an item breakdown on repair tooltip

Road-block:
Changelog  
api: GetColorGradient method to return a color c between color a and b given a ratio
panel: show damaged items on the repair tooltip with damage breakdown

Comments  
Some of the variables needed to be declared outside of the closure or they get "snapshotted" the first time the function returns [Ref:Closures](https://www.lua.org/pil/6.1.html)  

This also allows us to remove the 2nd loop over the inventory slots to get the repair cost for a small performance optimization.  

Credit to Lichery for starting the discussion with #905

### DIFF
--- a/api/api.lua
+++ b/api/api.lua
@@ -786,3 +786,27 @@ function pfUI.api.GetColoredTimeString(remaining)
     return ""
   end
 end
+
+-- [ GetColorGradient ] --
+-- 'perc'     percentage (0-1)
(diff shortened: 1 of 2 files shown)


Road-block:
new repair tooltip

Road-block:
Updated ColorGradient method so it can take more colors to transition through.  
Makes it possible to have a Green > Yellow > Red scale or even more colors instead of just a start and end color.

PR #912, opened 2018-05-07
Various Loot module improvements

Road-block:
**Changelog**  
loot: make detailed random roll announce an option, default: off
api: SendChatMessageWide will use raid warning if player is raid leader or assist
loot: rearrange advanced masterloot menus so random rolling and special recipients precede class menu
loot: fix a special recipients unitpopup menu check so they won't show in the context menu when the player is solo
loot: change the auto confirm bind when solo implementation, so player doesn't have to double-click the loot; also make sure it works if the loot bind confirmation staticpopup is not at index 1

**Comments**  
When I originally put in the random loot assign it was under the assumption that it will be used sparingly and in a PuG setting where announcing the full list of player and assigned number before doing the /roll and distributing is essential for trust.  
Since then quite a few people have reached out to tell me that it is also used as an alternative to auto-rollers for stuff like bijou / coins / scarabs etc so RLs can keep ML on the whole time and have someone just quickly random stuff.  
That makes the announce spam a problem.  
Additionally it's inconvenient for the person using the special menus frequently to have to traverse past the class menu to get to those options.  
So I re-arranged the menu to make random + special recipients more accessible and made the full random list announce an option.  
With the spam reduced or eliminated there was no more reason to not make SendChatMessageWide true to its name so it actually sends to the highest channel available according to player role.  

I also changed the implementation of the auto confirm BoP when solo because the previous one was forcing the player to double click the item to pick it up (the problem was two-fold: the static popup is shown by the default client through a registration in UIParent not LootFrame and that was never touched, pfUI was in addition trying to show the same static popup at the same event which due to it having the `exclusive = 1` flag set caused the UIParent one to hide and in essence clicked the 3rd static popup to show when the player clicked the loot slot again)

### DIFF
--- a/api/api.lua
+++ b/api/api.lua
@@ -212,7 +212,11 @@ end
 function pfUI.api.SendChatMessageWide(msg)
   local channel = "SAY"
   if UnitInRaid("player") then
+    if ( IsRaidLeader() or IsRaidOfficer() ) then
+      channel = "RAID_WARNING"
+    else
     channel = "RAID"
+    end
   elseif UnitExists("party1") then
     channel = "PARTY"
   end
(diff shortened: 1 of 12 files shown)


Lichery:
how did you make the new PR? I use the Github desktop to create an PR and in it I can add new commits only to the existing PR ...

Lichery:
lol I understood. You use different branches...

PR #924, opened 2018-05-19
Dropdown option in raid to set a player as tank

xergio:
When you select someone as tank, he will be placed at the beginnig of his group, usefull for heals to see all tank in the same row/column.

### DIFF
--- a/modules/raid.lua
+++ b/modules/raid.lua
@@ -39,30 +39,30 @@ pfUI:RegisterModule("raid", function ()
     end
   end
 
+  -- add units to the beginning of their groups
+  function pfUI.uf.raid:AddUnitToGroup(index, group)
+    for subindex = 1, 5 do
+      ids = subindex + 5*(group-1)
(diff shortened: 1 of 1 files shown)


xergio:
Feature in action: https://streamable.com/qvzxy

modules/raid.lua
```diff
@@ -32,29 +39,63 @@ pfUI:RegisterModule("raid", function ()
...
-  pfUI.uf.raid:RegisterEvent("VARIABLES_LOADED")
-  pfUI.uf.raid:SetScript("OnEvent", function()
+  function pfUI.uf.raid:RaidSetup()
     for i=1, 40 do
       pfUI.uf.raid[i].id = 0
       pfUI.uf.raid[i]:Hide()
     end
 
+    -- add units to the beginning of their groups
+    local function AddUnitToGroup(index, group)
```
shagu: It's a bad idea to define a function within another function that is called frequently. Please move it out of there.

xergio: done, PR updated

PR #941, opened 2018-05-31
pettarget: do not load module if pet frame is disabled

dein0s:
got this error after updating to latest master:
error_screen

as i use LUF - i've disabled all frame modules


Road-block:
Just a heads up you have misspelled `nil` as `nill`  
In many cases this will work the same as it's unlikely someone has defined a variable named `nill` but still ..

dein0s:
yea, thanks - fixed :) 
too late in the night after almost whole day of coding.

shagu:
https://github.com/shagu/pfUI/commit/532a212871c49721a601c1cb34b59fa0a770ae99

PR #949, opened 2019-12-25
Target of Target of Target

attackpure100:
don't know if this already exists or similar is in progress, but here it is. I use it for myself so thought i would add.

### DIFF
--- a/api/config.lua
+++ b/api/config.lua
@@ -712,7 +712,7 @@ function pfUI:MigrateConfig()
 
   -- migrating old to new unitframe texts (> 3.0.0)
   if checkversion(3, 0, 0) then
-    local unitframes = { "player", "target", "focus", "group", "grouptarget", "grouppet", "raid", "ttarget", "pet", "ptarget", "fallback", "tttarget" }
+    local unitframes = { "player", "target", "focus", "group", "grouptarget", "grouppet", "raid", "ttarget", "pet", "ptarget", "fallback" }
 
     for _, unitframe in pairs(unitframes) do
       if pfUI_config.unitframes[unitframe].txtleft then
@@ -758,7 +758,7 @@ function pfUI:MigrateConfig()
 
   -- migrating legacy buff/debuff naming (> 3.5.0)
   if checkversion(3, 5, 0) then
-    local unitframes = { "player", "target", "focus", "group", "grouptarget", "grouppet", "raid", "ttarget", "pet", "ptarget", "fallback", "tttarget" }
+    local unitframes = { "player", "target", "focus", "group", "grouptarget", "grouppet", "raid", "ttarget", "pet", "ptarget", "fallback" }
 
     for _, unitframe in pairs(unitframes) do
       local entry = pfUI_config.unitframes[unitframe]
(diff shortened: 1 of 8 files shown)


api/config.lua
```diff
@@ -700,7 +712,7 @@ function pfUI:MigrateConfig()
 
   -- migrating old to new unitframe texts (> 3.0.0)
   if checkversion(3, 0, 0) then
-    local unitframes = { "player", "target", "focus", "group", "grouptarget", "grouppet", "raid", "ttarget", "pet", "ptarget", "fallback" }
+    local unitframes = { "player", "target", "focus", "group", "grouptarget", "grouppet", "raid", "ttarget", "pet", "ptarget", "fallback", "tttarget" }
```
shagu: Migrations are not needed, as there was no tttarget in previous versions. The default you defined will be used anyways.

api/config.lua
```diff
@@ -746,7 +758,7 @@ function pfUI:MigrateConfig()
 
   -- migrating legacy buff/debuff naming (> 3.5.0)
   if checkversion(3, 5, 0) then
-    local unitframes = { "player", "target", "focus", "group", "grouptarget", "grouppet", "raid", "ttarget", "pet", "ptarget", "fallback" }
+    local unitframes = { "player", "target", "focus", "group", "grouptarget", "grouppet", "raid", "ttarget", "pet", "ptarget", "fallback", "tttarget" }
```
shagu: In this specific case, it makes sense, as it will adapt old buff settings to all new unitframes.

shagu: This also seems to be safe to removed, the `if entry.buffs` will never catch, as there won't be `pfUI_config.unitframes[unitframe]` where unitframe is `tttarget`.

api/config.lua
```diff
@@ -757,7 +769,7 @@ function pfUI:MigrateConfig()
 
   -- migrating glow settings (> 3.5.1)
   if checkversion(3, 5, 0) then
-    local common = { "player", "target", "ttarget", "pet", "ptarget"}
+    local common = { "player", "target", "ttarget", "pet", "ptarget", "tttarget"}
```
shagu: That's a useful one, keep it :)

api/config.lua
```diff
@@ -798,7 +810,7 @@ function pfUI:MigrateConfig()
 
   -- migrate buff/debuff position settings (> 3.19)
   if checkversion(3, 19, 0) then
-    local unitframes = { "player", "target", "focus", "group", "grouptarget", "grouppet", "raid", "ttarget", "pet", "ptarget", "fallback" }
+    local unitframes = { "player", "target", "focus", "group", "grouptarget", "grouppet", "raid", "ttarget", "pet", "ptarget", "fallback", "tttarget" }
```
shagu: As in previous versions there was no `TOP` or `BOTTOM` settings for tttarget, there is also nothing to be migrated here. Seems to be safe for removal.

attackpure100:
made changes on (fed05788d2550d12ee8308f353744cb99653223a)

shagu:
Thanks, you PR has been squashed and merget by this commit: https://github.com/shagu/pfUI/commit/050790b7e974c9bf1ccc1748cf78d7198845921d

(PS: I made a minor adjustment and disabled the frame by default `visible = 0`)

PR #968, opened 2020-06-14
map

elenmerbau:
add Mapster to the whitelist.


shagu:
This would disable the map module for everyone.

PR #981, opened 2020-07-13
Fix questitem enchanting

Retinus:
Fixes a bug when you hover mouse cursor over your enchanting recipes. With other professions didn't get any errors.


shagu:
Thank you :)

PR #984, opened 2020-07-25
Update translations_ruRU.lua

shikulja:
"Disable Errors in UIErrors Frame"
"Disable UIErrors Frame" -  would look better


shagu:
> "Disable UIErrors Frame" - would look better

I know, but it's not what it is. The UIErrorsFrame is used to display Errors such as "Out of Range" and Informations such as "Murloc slain 1/7". This option does only disable the errors.

shikulja:
I thought something like that it might be.  clear

shikulja:
the translation is correct, literal.

PR #986, opened 2020-07-28
xpbar: Added the option to not overlap the xpbar's rested bar to allow better color adjustment

DennisWG:
Currently, the xp bar's rested bar overlaps the actual xp bar which means that it is impossible to have a non-transparent rested color (as it would otherwise hide the xp bar [see images 1 and 2]). It also messes with the xp bar's color, as the rested bar's color will always be added ontop of it.

This pull request fixes this by adding a new option to push the rested bar below the xp bar, making only the actual rested part visible at the tail end of the player's current xp (see image 3). It is added as an option in order to not break already configured Interfaces and defaults to off.

Image 1 - The rested bar's transparency must always be below 100%

Image 2 - Transparency with 100% hides the xp bar

Image 3 - The bar with the new option enabled and a non transparent rested bar

I would appreciate any feedback!

Cheers

### DIFF
--- a/api/config.lua
+++ b/api/config.lua
@@ -545,6 +545,7 @@ function pfUI:LoadConfig()
   pfUI:UpdateConfig("panel",      "xp",          "rep_mode",         "VERTICAL")
   pfUI:UpdateConfig("panel",      "xp",          "rep_anchor",       "pfChatRight")
   pfUI:UpdateConfig("panel",      "xp",          "rep_position",     "LEFT")
+  pfUI:UpdateConfig("panel",      "xp",          "dont_overlap",         "0")
 
   pfUI:UpdateConfig("castbar",    "player",      "hide_blizz",       "1")
   pfUI:UpdateConfig("castbar",    "player",      "hide_pfui",        "0")
--- a/env/translations_deDE.lua
+++ b/env/translations_deDE.lua
@@ -643,6 +643,7 @@ pfUI_translation["deDE"] = {
   ["Standard Text Font Size"] = nil,
   ["Status"] = nil,
   ["Status Bar Texture"] = nil,
+  ["Don't overlap rested"] = nil,
   ["Switch Pages On Druid Stealth"] = nil,
   ["Switch Pages On Meta Key Press"] = nil,
   ["Systeminfo"] = nil,
--- a/env/translations_enUS.lua
+++ b/env/translations_enUS.lua
@@ -643,6 +643,7 @@ pfUI_translation["enUS"] = {
   ["Standard Text Font Size"] = nil,
   ["Status"] = nil,
   ["Status Bar Texture"] = nil,
+  ["Don't overlap rested"] = nil,
   ["Switch Pages On Druid Stealth"] = nil,
   ["Switch Pages On Meta Key Press"] = nil,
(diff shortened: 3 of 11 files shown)


api/config.lua
```diff
@@ -545,6 +545,7 @@ function pfUI:LoadConfig()
   pfUI:UpdateConfig("panel",      "xp",          "rep_mode",         "VERTICAL")
   pfUI:UpdateConfig("panel",      "xp",          "rep_anchor",       "pfChatRight")
   pfUI:UpdateConfig("panel",      "xp",          "rep_position",     "LEFT")
+  pfUI:UpdateConfig("panel",      "xp",          "dont_overlap",         "0")
```
shagu, line 548: except for the 4 spaces that indent the "0" to much, everything is fine :see_no_evil:

shagu:
Thanks for your contribution. I like it a lot.

shagu:
Your contribution has been merged by this commit: https://github.com/shagu/pfUI/commit/09498bc6cbd0da859fbb9c6335fd7bdfeb9b5e4b

shagu:
> I would appreciate any feedback!

**translations:**
You don't have to necessarily include the translations it in the PR if it is too much work doing it manually. I have a script that automatically generates the missing entries and I run that from time to time. So things would be added automatically a few days later. However, feel free to add them as it makes things easier, just saying that you don't have to :)

**commit message:**
I always try to keep the header limited to a maximum of 50 characters. If its slightly more, it is okay. But 50 is the reference.
If there is more to tell, use the body for a detailed description. Also I'm writing commit messages not about "what this has done" but about "what it will do".

`xpbar: added the option to foo` -> `xpbar: add an option to foo`

api/config.lua
DennisWG: Oh no! This one actually hurts D:
 Will fix this in another PR!

shagu: I already fixed your commit :D

DennisWG:
```
translations:
You don't have to necessarily include the translations it in the PR if it is too much work doing it manually. I have a script that automatically generates the missing entries and I run that from time to time. So things would be added automatically a few days later. However, feel free to add them as it makes things easier, just saying that you don't have to :)
```

I was thinking about adding the german translation when I realised that pfUIhasn't been localized into german at all. So I opted into not doing it either, since it would have looked out of place. But I don't adding them manually.

```
commit message:
I always try to keep the header limited to a maximum of 50 characters. If its slightly more, it is okay. But 50 is the reference.
If there is more to tell, use the body for a detailed description. Also I'm writing commit messages not about "what this has done" but about "what it will do".

xpbar: added the option to foo -> xpbar: add an option to foo
```

Got it! Will keep this in mind for future PRs (and I have some ideas for one or two more things) ;-)

shagu:
> I was thinking about adding the german translation when I realised that pfUI hasn't been localized into german at all. So I opted into not doing it either, since it would have looked out of place. But I don't adding them manually.

Feel free to do the translation work :-) I never found the time doing so. Plus I find it also quite hard to translate thing that you're already familiar with. Like should terms like "Unit Frames" be translated? I'm still hoping that someone will do it.. someday.

PR #991, opened 2020-08-18
remove percent when displaying energy/rage/focus

bidaumm:
In powerdyn / "Mana - Auto" mode removes the percentage if mpmax is < 101 (i.e a energy/rage/focus bar) which can be off by 1 digit at times and look confusing (i.e energy looks like "35 - 36%")
also spaces the energy num display two spaces to the right for easier identification due to smaller number vs mana and clipping against rested icon

### DIFF
--- a/api/unitframes.lua
+++ b/api/unitframes.lua
@@ -2179,7 +2179,10 @@ function pfUI.uf:GetStatusValue(unit, pos)
       return unit:GetColor("power") .. pfUI.api.Abbreviate(power)
     end
   elseif config == "powerdyn" then
-    if mp ~= mpmax then
+    -- dont show percentage for rage/energy/focus
+    if mpmax < 101 then
+      return unit:GetColor("power") .. "  " .. pfUI.api.Abbreviate(mp)
+    elseif mp ~= mpmax then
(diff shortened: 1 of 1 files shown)


bidaumm:
also spaces the energy display two spaces to the right for easier identification due to smaller number vs mana

shagu:
> "mpmax == 100" wouldn't return true

I don't understand why mpax == 100 wouldn't return true :thinking: 
I have never seen UnitManaMax returning 99 for rage/focus/energy.

shagu:
> also spaces the energy num display two spaces to the right for easier identification due to smaller number vs mana and clipping against rested icon

I don't see the reason. Do you have a screenshot?

bidaumm:
Replaced with mpax == 100  again and its working this time, I must have been doing something wrong.
Also removed the white spaces as per discord convo

bidaumm:
replaced mpmax==100 with  UnitPowerType(unitstr) > 0

shagu:
Thanks! Squashed and merged by this commit: https://github.com/shagu/pfUI/commit/bc4caea63c44d6c69399a16ec3a7284fd86ff472

PR #1035, opened 2021-03-22
Fix x and y coord update in frame unlock popup.

Maczuga:
Example: https://cdn.discordapp.com/attachments/695634717613228043/823649632633749514/unknown.png

You cannot actually edit those boxes. There are 2 things being wrong with it:
1. It awaits an "Escape" button press to modify value, seems counterintuitive, addons like ElvUI do that on Enter press.
2. It does not update values. SetPoint uses previous values, new value is never being used.

This fixes the boxes and mimics the ElvUI behavior for Enter button being the "save" action.


shagu:
Thanks for your contribution :+1:

PR #1075, opened 2022-08-03
unitframes: add power bar texture

mrrosh:
add an option to allow selection of the power bar texture
Screenshot (2)


shagu:
I have added config migrations, so that those who are updating, automatically get the health bar texture set as the initial entry for the power bar texture: 
https://github.com/shagu/pfUI/commit/28170cc5b7b4a0b9d95cf1a9658f0da3baacc168

And also updated the profiles here:
https://github.com/shagu/pfUI/commit/aa244b15ee110fe70a0ee74dc0d1146a155b1f47

PR #1079, opened 2022-08-22
Change druid bar texture & Add tooltip font and font size option

mrrosh:
Change DruidManaBar texture to the same one used by the power bar.


shagu:
Are you sure? I don't think it should be in quotes

mrrosh:
If I leave it outside quotes it resets the bar to where the Blizzard unit frames would be and doesn't change the texture.

PR #1088, opened 2022-09-09
[WIP] WOTLK Support

BenDol:
Very experimental and incomplete. Any suggestions or improvements are welcome!

I'm not sure if this is even in scope for pfUI at this stage, but figured since I would like it for WOTLK that I'd track its progress here.

image

Thanks!


pfUI.toc
```diff
@@ -1,9 +1,8 @@
-## Interface: 11200
-## Title: |cff33ffccpf|cffffffffUI
+## Interface: 30300
+## Title: |cff33ffccpf|cffffffffUI |cffaa5555[incomplete]
 ## Author: Shagu
 ## Notes: A complete user interface replacement.
-## Notes-ruRU: Полная замена пользовательского интерфейса.
```
shikulja, line 5: Don't need touch

BenDol: We can ignore this files changes for now, I'll be syncing it with the main file

PR #1096, opened 2022-11-26
libs: update kidney shot timer and return correct rank

gl00m3r:
Game client always return nil value with open quotes but with empty quotes it does works correct.

The Rogue has 2 ranks of Kidney shot — Rank 1 = 1s per combopoint and Rank 2 increase the time to 2s (+1s) for first combopoint so in total:
- Rank 1 (5 combopoints) = 5s duration
- Rank 2 (5 combopoints) = 6s duration

[Source](https://vanilla-wow-archive.fandom.com/wiki/Kidney_Shot)

### DIFF
--- a/libs/libdebuff.lua
+++ b/libs/libdebuff.lua
@@ -28,7 +28,7 @@ local lastspell
 
 function libdebuff:GetDuration(effect, rank)
   if L["debuffs"][effect] then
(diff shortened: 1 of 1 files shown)


gl00m3r:
I think the better why just change the locales.

PR #1097, opened 2022-11-26
Fix kidney shot duration and return correct rank of spell

gl00m3r:
Game client always return nil value with open quotes but with empty quotes it does works correct.

The Rogue has 2 ranks of Kidney shot — Rank 1 = 1s per combopoint and Rank 2 increase the time to 2s (+1s) for first combopoint so in total:
- Rank 1 (5 combopoints) = 5s duration
- Rank 2 (5 combopoints) = 6s duration

[Source](https://vanilla-wow-archive.fandom.com/wiki/Kidney_Shot)


shagu:
Thanks for your PR. I don't see why the last change is required. The rank that is passed to the function is a string like: `Rank 1`. The "RANK" constant that gets removed is `Rank`. So what remains is ` 1` with a leading whitespace.
For example if I add some debugging here:
Bildschirmfoto vom 2022-11-26 18-42-37
It will look ingame like this:
Bildschirmfoto vom 2022-11-26 18-31-41

Which locale are you playing on? Because for enUS it is perfectly fine.

gl00m3r:
It was a mistake on my part not to specify the input data:
- I play on [Everlook](Everlook.org) server (vanilla 1.12 VMangos Core)
- I use English language on **pfUI** so it is **enUS** locale I guess

I spent a few hours yesterday to understand why my little fix for **Kidney Shot** doesn't work at all. I used **SuperMacro** for tests. I tried todo something like this:
```lua
elseif effect == L["dyndebuffs"]["Kidney Shot"] then
      if rank == 1 then duration = GetComboPoints()*1
      else duration = duration + GetComboPoints()*1 end
```
So I was start digging and I knew about `RANK`.

First I was want to see what value `rank` has at `local rank = rank and tonumber((string.gsub(rank, RANK .. " ", ""))) or 0` line.  I made a macro `tonumber((string.gsub(rank, RANK .. " ", "")))` it returns `nil` value. You should be ask: `why?`. I was confusing too. I tried to delete whitespace `RANK .. ""` and it fix that :)

Today I figured out why that happened. After tests on an other character with only turned **pfUI** and **SuperMacro** it works as expected, return correct value: `1`. So weird. That behavior causes by the **LazyScript** addon exactly **LazyRogue** extension for rogue class. Probably his rest extensions cause the same issue.

Probably you are right and https://github.com/shagu/pfUI/pull/1097/commits/28ee2c48bf7dd9f53ec0415ecf6b77c4165db463 changes unnecessary but it fix some weird compatibles.

PR #1101, opened 2022-12-24
Castbar: Add options to hide spell names and timers

Pizzahawaiii:


env/profiles.lua
```diff
@@ -167,6 +167,8 @@ f:SetScript("OnEvent", function()
     },
     ["castbar"] = {
       ["player"] = {
+        ["showname"] = "1",
+        ["showtimer"] = "1",
```
shagu, line 171: Those are not required. If something doesn't exist in the profile, it will be set to the default value, which is already 1 :)

modules/castbar.lua
```diff
@@ -45,7 +45,6 @@ pfUI:RegisterModule("castbar", "vanilla:tbc", function ()
     cb.bar.left:SetFontObject(GameFontWhite)
     cb.bar.left:SetTextColor(1,1,1,1)
     cb.bar.left:SetFont(font, font_size, "OUTLINE")
-    cb.bar.left:SetText("left")
```
shagu, line 48: good catch :see_no_evil:

modules/castbar.lua
```diff
@@ -57,7 +56,6 @@ pfUI:RegisterModule("castbar", "vanilla:tbc", function ()
     cb.bar.right:SetFontObject(GameFontWhite)
     cb.bar.right:SetTextColor(1,1,1,1)
     cb.bar.right:SetFont(font, font_size, "OUTLINE")
-    cb.bar.right:SetText("right")
```
shagu, line 60: Guess i forgot some debug stuff :see_no_evil: 
Thanks

env/profiles.lua
Pizzahawaiii: Ah nice, makes sense. Thanks! 🙌

Pizzahawaiii:
Thanks for the quick review! 🙌 

I wasn't sure if these options were left out on purpose and if I should contribute them to upstream. But I'm sure they have their uses for people. E.g. I'm working on an ultra-minimalistic UI (see screenshot) where I'm trying to get rid of everything that's not absolutely necessary. I can still tell which spell my taret is casting from the icon and obviously, I always know what I'm casting myself, so the spell names aren't needed.

What's your opinion @shagu, should this be merged?

WoWScrnShot_122422_165254

shagu:
> What's your opinion @shagu, should this be merged?

Sorry I was a busy during christmas and only did a quick review on mobile.
I never merge on my phone so this had to wait :)

> Also, merry christmas! santachristmas_treegift
Thank you, Merry christmas!

PR #1131, opened 2023-03-10
Turtle WoW Skins

doorknob6:
Adds skins for the custom turtle wow character and inspect talent frames.

### DIFF
--- a/init/skins.xml
+++ b/init/skins.xml
@@ -46,4 +46,5 @@
 
   <!-- Turtle -->
   <Include file="..\skins\turtle\character.lua"/>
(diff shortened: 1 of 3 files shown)


shagu:
Thank you for your contribution, but I'd like to avoid custom server stuff in the main repository where possible.
I'd prefer to have it as a seperate addon, that adds the turtle-wow specific changes (like cast times, debuff timers, skins, talent scanner changes, etc). For example, this is a template that was used by some players of vanilla-plus to add their required custom changes: https://github.com/shagu/pfUI-vanillaplus

doorknob6:
Alright, figured that might be the case. I will look into it.

PR #1134, opened 2023-03-12
Pull who information for players with unknown classes

xIGBClutchIx:
Same as pull request #1132 just added a new branch for local development of features.

This allows for the saving of class/level information for unknown players in other chats like global. So on the next chat message, they will actually have a class color.

This patch is a bit hacky but seems to be fine and been testing and have yet to encounter any issues. Got the idea from CleanChat.


shagu:
I had something similar on a local branch for a while. But never made it to the main repository, because it broke on several servers. I remember on Kronos launch, the social window became completely defunct as they time-limited the /who queries. On another realm i got disconnects for sending too many /who-queries on peak hours.

So I can't merge it as is. But I'd be fine if you add an option for it, which is disabled by default.

xIGBClutchIx:
Sounds good, I just added an option for that with disable by default.

As for the pattern, I guess we can add that into the locales but I was assuming those were generated somehow?

shagu:
Some are auto-generated, some are manual. However, I don't think it is needed at all.
I used things like that on my branch:
```lua
local nothing = function() return end
local original = FriendsFrame_OnEvent

local who_query = CreateFrame("Frame")
who_query:RegisterEvent("WHO_LIST_UPDATE")
who_query:SetScript("OnEvent", function()
  if this.pending then
    -- restore everything once a query is received
    _G.FriendsFrame_OnEvent = original  
    this.pending = nil
    SetWhoToUI(0)
  end
end)

local scan_name = function(name)
  -- abort if another query is ongoing
  if who_query.pending then return end
  who_query.pending = true

  -- prepare and send the who query
  _G.FriendsFrame_OnEvent = nothing
  SetWhoToUI(1)
  SendWho("n-"..name)
end
```

xIGBClutchIx:
Love it so much cleaner and easier

PR #1182, opened 2023-07-18
[WIP] Nameplates: Low Health, Name Reaction Colors

ghost:
A couple of ideas which might be useful to add to the nameplates.

Low Health Colors will color the health bar when at 20% health or lower (different color for friendlies and enemies)
Name Reaction  Colors will color the unit's name by reaction (helpful when using it with the class color health bar option)

The code could possibly be improved but thought i'd leave this here as a proof of concept/suggestion!

Thanks!

lowhealthc

namereactionc


shagu:
Hi,
Thanks for the PR. I like the idea of the low-health display, but I would prefer it to have it begin to slowly fade to red when the unit goes below 50%, but I can't be bothered to do the math right now :D 

Regarding the name colors, I see the need, but it also has the downside to no longer see the infight notifications.
Personally I'm fine with it, but I already see the questions incoming why the infight name color change is no longer visible :see_no_evil:

GryllsAddons:
Thanks for the review!

I really like your idea of fading, I'll leave that to you as the colour math goes over my head! 😄 it would be great to add something as it's really useful to know at a glance when people/mobs are low or can be executed.

With the name colors it's a difficult one! I can see the value of the infight colors and it would be a shame to lose that piece of information. I've been trying to think of ideas of how to have both reaction and combat info visible at a glance while keeping the frames simple and uncluttered. 

One idea I thought of was to include an option for a combat icon option. I've uploaded a commit (see pic below) that will show/hide the icon/level while in combat.

I'll have a go at some other solutions as well and perhaps find something that works well to blend both!

image

PR #1247, opened 2023-12-24
Add TW Threatmeter integration

Zedorff:
Added integration for [TWThreatMeter](https://github.com/balakethelock/TWThreat) into right chatbar
Supported both single and double variants


Zedorff:
image

PR #1249, opened 2024-01-13
Add all glyphs of FRIZQT__.TTF to PFUI fonts +  raid market unicodes

balakethelock:
Merged base game's FRIZQT__.TTF font to all the missing glyphs from the PFUI fonts, this makes it compatible with addons that use glyphs in their interfaces. In addition, I added some new glyphs that don't exist in Frizqt to represent the raid marker icons and this is how they look ingame: 
image

The difference: screenshot taken using the website [FontDrop](https://fontdrop.info/#/?darkmode=true)
Before
image
After
image


shagu:
Thank you for your contribution. After having this laying around for way too long, I had to make a decision finally.
I have decided to not merge it due to compatibility and legal reasons. Some of the fonts are already in the very grey area of being used. Modifying those would make it even worse. I really like the fonts and would want to keep them, so I think it's better to at least keep them as the creator of them intended them to be.

You could still create an addon like `pfUI-fonts` to exchange the fonts on your own.
Sorry for the long time without any decisions being made, and thanks again for your work.

PR #1257, opened 2024-02-15
Don't rangecheck when player has combo points to avoid losing them

gwetchen:


shagu:
Hi, thanks for your contribution, 

I think it could be done easier like that:
```lua
local combo = CreateFrame("Frame", "pfComboPointsDetect")
combo:RegisterEvent("PLAYER_COMBO_POINTS")
combo:SetScript("OnEvent", function()
  PlayerFrame.combopoints = GetComboPoints() > 0
end)
```

Also I'm not sure if `.combopoints` on a PlayerFrame might be to generic, I could imagine this turning into a conflict with other addons. I'd prefer to play safe and just use a local variable to save the value to.

PS: there's a typo in the comment :rabbit:

gwetchen:
Thank you for the feedback, I really appreciate it. 
I hope this is better.

I've removed the 🐰:)

shagu:
Thanks, merged :)

PR #1259, opened 2024-02-21
Added possibility to change nameplates height

ne0nko:
Hi there. It is very useful to have the ability to change the height of the nameplate. I don't know why it is missing in the menu, so I fixed it.
image


shagu:
image

Issue #1127, opened 2023-02-21
ISsue with getting buttons to bind


cyemekaray:
I cannot assign the mouse wheel up, mouse wheel down, mouse button 3 and other mouse buttons, using Hoverbind. Also I cannot press buttons on panels using any mouse buttons other than left and righ.

shagu:
The hoverbind feature does only support keys. 
If you need to bind mouse buttons, you would need to use the regular Keybind menu of the game.
Whenever I decide to work on that, it will be tracked in this issue: https://github.com/shagu/pfUI/issues/1016

Issue #1016, opened 2020-12-24
Buttons binding


CTPEJIOK84:
I cannot assign the mouse wheel up, mouse wheel down, mouse button 3 and other mouse buttons, using Hoverbind. Also I cannot press buttons on panels using any mouse buttons other than left and righ.

shagu:
Fixed via https://github.com/shagu/pfUI/commit/ba5e0abecf8599e7093e16af5a40990ce953574f

PR #1269, opened 2024-03-21
hoverbind: add support for mouse button & mouse wheel bindings

Pizzahawaiii:
Fixes 🐭 #1016 and 🐁 #1127

# Summary

This extends hoverbind mode to allow binding mouse buttons and mouse wheel up/down to actionbar buttons. It does this by overlaying all actionbar buttons with separate invisible frames that receive all mouse and keyboard events and handle binding creation and removal.

# Versions Tested

Tested and everything seems to work fine on the following clients:

- 1.12.1 (5875)
- 1.17.1 (7100) (TurtleWoW)
- 2.4.3 (8606)

# Notes/Questions

1. Against my own expectations, I actually managed to make this work somehow... 😄 Thanks again for the ingame support last night @shagu! ❤️
2. Since I'm new to addon development, I'm not sure if this is the most elegant implementation. I'll let you be the judge and I'm looking forward to any kind of feedback.
3. ~~Do we even need the `need_save` variable & logic? The code would be nicer and simpler if we just always blindly `SaveBindings()`, even if there are no changes. Would that cause any (e.g. performance) issues?~~
**Update:** I removed this so we always just save bindings immediately. ([839a718](https://github.com/shagu/pfUI/pull/1269/commits/839a718bd5e6efc433515ce0ca8c5d304c8b018d))

### DIFF
--- a/modules/actionbar.lua
+++ b/modules/actionbar.lua
@@ -1464,6 +1464,8 @@ pfUI:RegisterModule("actionbar", "vanilla:tbc", function ()
   pfUI.bars.update = updatecache
   pfUI.bars.buttons = buttoncache
   pfUI.bars.ButtonFullUpdate = ButtonFullUpdate
+  pfUI.bars.ButtonEnter = ButtonEnter
+  pfUI.bars.ButtonLeave = ButtonLeave
 
   pfUI.bars.UpdateGrid = function(self, state, typ)
     if not typ then
--- a/modules/hoverbind.lua
+++ b/modules/hoverbind.lua
@@ -36,6 +36,12 @@ pfUI:RegisterModule("hoverbind", "vanilla:tbc", function ()
     ["SHIFT"] = "SHIFT-"
   }
 
+  -- We don't allow binding these keys without any modifiers
+  local blockedKeys = {
+    "LeftButton",
+    "RightButton",
+  }
+
   pfUI.hoverbind = CreateFrame("Frame","pfKeyBindingFrame",UIParent)
   pfUI.hoverbind:Hide()
   pfUI.hoverbind:RegisterEvent("PLAYER_REGEN_DISABLED")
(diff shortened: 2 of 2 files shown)


shagu:
This is great! Thank you very much for your contribution. :+1: 
I won't have much time today. But I will test and review your PR as soon as possible.

shagu:
I have reviewed and tested it. It's looks very good, there is only one thing missing that I'd like to see:

I'd say there should be a safeguard to prevent users binding the Left-Mousebutton by accident.
The Left-Mousebutton is by default set to interact/action, which is a hidden binding in the 1.12.1 client:

```
  <!-- Hidden bindings -->
  <Binding name="TURNORACTION" runOnUp="true" hidden="true">
    if ( keystate == "down" ) then 
      TurnOrActionStart();                                                                                                                                                                 
    else 
      TurnOrActionStop();
    end  
  </Binding>
```

That means, people can't use the default keybind window to undo their change if they misclicked.
The only option they'd have is to ask for support and then to macro their way out of the misery.
(Or reset ALL keybindings and start over again)

In my opinion, checking in hoverbind-code for a normal click (without modifiers) and ignoring the event would be the best solution here.

Pizzahawaiii:
Thanks for the review!

You're right, we definitely don't want that. The thought had crossed my mind that binding left (or right) mouse button is kind of pointless. But I wasn't aware of the fact (and didn't consider the possibility) that binding left mouse button would render the player unable to use the UI, including the default menu, so I just left it there. Good catch!

Do you think we should also block the right mouse button in the same fashion, or merge as-is and leave that up to the player?

shagu:
> Do you think we should also block the right mouse button in the same fashion, or merge as-is and leave that up to the player?

Removing both left and right button sounds good :)

Pizzahawaiii:
Alright, done! ✅

shagu:
Awesome! Good job! Merged :)

Issue #1283, opened 2024-05-31
libpredict.lua: align and optimize the way the hooked functions "CastSpell", "CastSpellByName" and "UseAction" set the unit-name on spell_queue[3]


dsidirop:
- (optimization) Places where we use UnitName("player") can be replaced by simply 'player'

- (bug-fix) We only take into account the pfui uf-mouseover-unit in 'CastSpellByName' post-hook but we could take it into account in the other two post-hooks as well to have all 3 implementations aligned

- (optimization) We could intelligently cache the pfUI.uf.mouseover object the first time we have to consult it in any of the hooks, so as to save ourselves some time in subsequent invocations

- (optimization) If pfUI.uf.mouseover.unit == "player" we can simply return the 'player' variable

- (optimization) In the hook of 'CastSpellByName' we can avoid calculating the 'default' value every time unless it's absolutely necessary

Will make a PR about this soon. Appreciate this awesome addon btw ;)

ksidirop-laerdal:
We can close this one as wont-fix/rejected @shagu

PR #1284, opened 2024-05-31
#1283 optimize and align libpredict spell cast hooks

dsidirop:
https://github.com/shagu/pfUI/issues/1283

### DIFF
--- a/libs/libpredict.lua
+++ b/libs/libpredict.lua
@@ -316,18 +316,24 @@ hooksecurefunc("CastSpell", function(id, bookType)
   spell_queue[3] = UnitName("target") and UnitCanAssist("player", "target") and UnitName("target") or UnitName("player")
 end, true)
 
+local _cached_pfui_uf_mouseover -- will get cached upon first use in one of the hooks below 
 hooksecurefunc("CastSpellByName", function(effect, target)
   if not libpredict.sender.enabled then return end
-  local effect, rank = libspell.GetSpellInfo(effect)
-  if not effect then return end
-  local mouseover = pfUI and pfUI.uf and pfUI.uf.mouseover and pfUI.uf.mouseover.unit
-  mouseover = mouseover and UnitCanAssist("player", mouseover) and UnitName(mouseover)
+  local effectRaw, rank = libspell.GetSpellInfo(effect)
+  if not effectRaw then return end
 
-  local default = UnitName("target") and UnitCanAssist("player", "target") and UnitName("target") or UnitName("player")
+  _cached_pfui_uf_mouseover = _cached_pfui_uf_mouseover or pfUI.uf.mouseover or {}
+  local pfui_uf_mouseover_unit = _cached_pfui_uf_mouseover.unit
 
-  spell_queue[1] = effect
-  spell_queue[2] = effect.. ( rank or "" )
-  spell_queue[3] = mouseover or default
+  spell_queue[1] = effectRaw
+  spell_queue[2] = effectRaw .. (rank or "")
+  spell_queue[3] = pfui_uf_mouseover_unit == "player"
+          and player
+          or (
+          (pfui_uf_mouseover_unit and UnitCanAssist("player", pfui_uf_mouseover_unit))
+                  and UnitName(pfui_uf_mouseover_unit) -- mouseover unit name
+                  or (UnitCanAssist("player", "target") and UnitName("target") or player) -- or default
+  )
 end, true)
 
 local scanner = libtipscan:GetScanner("prediction")
--- a/libs/libpredict.lua
+++ b/libs/libpredict.lua
@@ -307,16 +307,28 @@ local function UpdateCache(spell, heal, crit)
 end
 
 -- Gather Data by User Actions
+
+local _cached_pfui_uf_mouseover -- will get cached upon first use in one of the hooks below
+
 hooksecurefunc("CastSpell", function(id, bookType)
   if not libpredict.sender.enabled then return end
-  local effect, rank = libspell.GetSpellInfo(id, bookType)
-  if not effect then return end
-  spell_queue[1] = effect
-  spell_queue[2] = effect.. ( rank or "" )
-  spell_queue[3] = UnitName("target") and UnitCanAssist("player", "target") and UnitName("target") or UnitName("player")
-end, true)
+  local effectRaw, rank = libspell.GetSpellInfo(id, bookType)
+  if not effectRaw then return end
 
-local _cached_pfui_uf_mouseover -- will get cached upon first use in one of the hooks below 
+  _cached_pfui_uf_mouseover = _cached_pfui_uf_mouseover or pfUI.uf.mouseover or {} 
+  local pfui_uf_mouseover_unit = _cached_pfui_uf_mouseover.unit
+
+  spell_queue[1] = effectRaw
+  spell_queue[2] = effectRaw .. (rank or "")
+  spell_queue[3] = pfui_uf_mouseover_unit == "player"
+          and player
+          or (
+          (pfui_uf_mouseover_unit and UnitCanAssist("player", pfui_uf_mouseover_unit))
+                  and UnitName(pfui_uf_mouseover_unit) -- mouseover unit name
+                  or (UnitCanAssist("player", "target") and UnitName("target") or player) -- or default
+  )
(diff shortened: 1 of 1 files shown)


shagu:
Hi,

What's the reason for `_cached_pfui_uf_mouseover` if it always points to pfUI.uf.mouseover` anyways?

I understand everything in this PR should optimize something, but all I see are more variables allocated and more logic running in each key press. Caching "player" totally makes sense to me, but what's the benefit of everything around it?

dsidirop:
Good questions.

> What's the reason for _cached_pfui_uf_mouseover if it always points to pfUI.uf.mouseover` anyways?

Old implementation:

```
local mouseover = pfUI and pfUI.uf and pfUI.uf.mouseover and pfUI.uf.mouseover.unit
```

- pfUI always exists so no point to check it over and over
- pfUI.uf always exists so no point to check it over and over

1st simplification:

```
local mouseover = pfUI.uf.mouseover and pfUI.uf.mouseover.unit
```

Moreover:

- pfUI.uf.mouseover only exists if the mouseover module is loaded (otherwise it's nil):

If the 'mouseover.lua' module is enabled it always gets loaded **after** 'libpredict.lua' so we cannot snapshot the 'pfUI.uf.mouseover' when we define '_cached_pfui_uf_mouseover' (outside the hooks) in 'libpredict.lua' because at that point in time 'pfUI.uf.mouseover' will **always** be 'nil' (ouch).

One solution to this problem is the one demonstrated in the patch: We delay snapshotting 'pfUI.uf.mouseover' when any of the 3 hooks gets triggered for the 1st time and in this way we're "sure" that we get the proper value of 'pfUI.uf.mouseover'.

And to make things a bit easier on us if the 'mouseover' module is not loaded (pfUI.uf.mouseover == nil) we can simply default '_cached_pfui_uf_mouseover ' to a dummy empty object '{}' as placeholder so as to simplify the conditional checks it gets involved in below (in which case the .unit will always be 'nil' which is what we want in this scenario).

All in all, the following doesn't generate any allocation apart from the very first time this line is executed in any of the 3 hooks:

```
_cached_pfui_uf_mouseover = _cached_pfui_uf_mouseover or pfUI.uf.mouseover or {}
```

From the second time onwards this will always always boil down to self-assignment which the lua engine can detect and handle very fast:

```
_cached_pfui_uf_mouseover = _cached_pfui_uf_mouseover 
```

- This does generate an allocation of a variable indeed:

```
  local pfui_uf_mouseover_unit = _cached_pfui_uf_mouseover.unit
```

We are employing the variable in up to 4 places and if we hadn't stored it in a local variable like above we would be forced to perform up to 4 table look ups (writing _cached_pfui_uf_mouseover.unit is similar to _cached_pfui_uf_mouseover["unit"]).

I have the feeling that storing the .unit to a local variable in the execution frame is far more performant than having 2/4 table-lookups but feel free to refactor this if you feel otherwise.

```
  spell_queue[3] = pfui_uf_mouseover_unit == "player"
          and player
          or (
          (pfui_uf_mouseover_unit and UnitCanAssist("player", pfui_uf_mouseover_unit))
                  and UnitName(pfui_uf_mouseover_unit) -- mouseover unit name
                  or (UnitCanAssist("player", "target") and UnitName("target") or player) -- or default
  )
```

- Optimizing the following logic:

```
mouseover = mouseover and UnitCanAssist("player", mouseover) and UnitName(mouseover)
local default = UnitName("target") and UnitCanAssist("player", "target") and UnitName("target") or UnitName("player")
[...]
spell_queue[3] = mouseover or default
```

Calculating the 'default' variable can be avoided altogether if 'mouseover' yields a truthy result like so:

```
mouseover = mouseover and UnitCanAssist("player", mouseover) and UnitName(mouseover)

spell_queue[3] = mouseover or (UnitName("target") and UnitCanAssist("player", "target") and UnitName("target") or UnitName("player"))
```

Lua is smart enough to stop the evaluation of the 'or' condition if 'mouseover' is truthy. We can optimize this further by inlining the whole thing:

```
spell_queue[3] = (mouseover and UnitCanAssist("player", mouseover) and UnitName(mouseover))  -- mouseover
           or (UnitName("target") and UnitCanAssist("player", "target") and UnitName("target") or UnitName("player"))  --default
```

We now notice that in the 'default' part we invoke UnitName("target") twice. We can remove the first invocation of it and the result will always be exactly the same:

```
spell_queue[3] = (mouseover and UnitCanAssist("player", mouseover) and UnitName(mouseover))  -- mouseover
           or (UnitCanAssist("player", "target") and UnitName("target") or UnitName("player"))  --default
```

And ofcourse we can use the cached 'player' name:

```
spell_queue[3] = (mouseover and UnitCanAssist("player", mouseover) and UnitName(mouseover))  -- mouseover
           or (UnitCanAssist("player", "target") and UnitName("target") or player)  --default
```

Finally we observe that 'mouseover' can be set to target "player" in which case we can guard close very early:

```
-- afaik such string-checks are blazing fast for lua as strings are interned https://www.luafaq.org/gotchas.html#T9
if mouseover == "player" then
      spell_queue[3] = player
      return
end

spell_queue[3] = (mouseover and UnitCanAssist("player", mouseover) and UnitName(mouseover))  -- mouseover
           or (UnitCanAssist("player", "target") and UnitName("target") or player)  --default
```

Same thing can be written with an inlined condition:

```
spell_queue[3] = mouseover == "player"
       and player
       or (
                 (mouseover and UnitCanAssist("player", mouseover) and UnitName(mouseover))  -- mouseover
           or  (UnitCanAssist("player", "target") and UnitName("target") or player)  --default
       )
```

Which is what we have in the PR. Maybe the if-check is a bit more clear and performant. Feel free to apply that approach as it's probably more readable too. Just my 2c.

Appreciate the great framework and thanks for everything!

shagu:
> Old implementation:
> 
> ```
> local mouseover = pfUI and pfUI.uf and pfUI.uf.mouseover and pfUI.uf.mouseover.unit
> ```
> 
>     * pfUI always exists so no point to check it over and over
> 
>     * pfUI.uf always exists so no point to check it over and over
> 
> 
> 1st simplification:
> 
> ```
> local mouseover = pfUI.uf.mouseover and pfUI.uf.mouseover.unit
> ```

This woul be correct for the pfUI space. But pfUI does not always exists and pfUI.uf don't either if the library is used outside the pfUI context. For example within ShaguTweaks-mods or other projects that already begun to use it.
The check is very conservative and iterates over the pfUI tree to not abort in error, or require others to patch out everything pfUI related.

The `and` checks of existance are very fast within Lua and I accepted it for the sake of not maintaining multiple versions of the library.

> Moreover:
> 
>     * pfUI.uf.mouseover only exists if the mouseover module is loaded (otherwise it's nil):
> 
> 
> If the 'mouseover.lua' module is enabled it always gets loaded **after** 'libpredict.lua' so we cannot snapshot the 'pfUI.uf.mouseover' when we define '_cached_pfui_uf_mouseover' (outside the hooks) in 'libpredict.lua' because at that point in time 'pfUI.uf.mouseover' will **always** be 'nil' (ouch).

I still fail to understand why we **need** to snapshot it at all? I know, accessing locals is faster than accesing an item from a tree, but it's not **that** much faster that it would be worth to bloat the code like this, is it?

> I have the feeling that storing the .unit to a local variable in the execution frame is far more performant than having 2/4 table-lookups but feel free to refactor this if you feel otherwise.

I don't have the feeling that accessing pfUI.uf.mouseover.unit on a keypress has a noticable impact at all.
Especially not in comparison to what else runs on that keypress (like setting a tooltip, reading over it, ...)

As much as I understand and appreciate your efforts here, the readability is far more important to me
than squeezing out like 0.000001s per keypress. In the end it will be me again who has to maintain it.
And the new version looks bloated and more confusing to my future self in 2 years than the current one.

> Feel free to apply it as it's probably more readable too. Just my 2c.
> 
> Appreciate the great framework and thanks for everything!

Thank you :blush: 

I hope you understand my reasons for declining it.

dsidirop:
Understood. Just a quick question:

> But pfUI does not always exists and pfUI.uf don't either if the library is used outside the pfUI context. For example within ShaguTweaks-mods or other projects that already begun to use it.

Are these libraries using pfui in a way that causes this file to be loaded and run without pfUI or pfUI.uf being loaded?

Issue #1285, opened 2024-05-31
Fix a bug which prevents libspell.GetSpellInfo() from deducing the proper min-range, max-range and casting-time


dsidirop:
This is a harmless bug as far as pfui itself is concerned because it doesn't use min-range, max-range and casting-time.

But this bug can hurt 3rd party addons that employ libspell.GetSpellInfo() so it would be nice to sort this out just in case.

PR #1287, opened 2024-05-31
#1285 fix libspell getspellinfo() for ranges and casting-times

dsidirop:
https://github.com/shagu/pfUI/issues/1285

### DIFF
--- a/libs/libspell.lua
+++ b/libs/libspell.lua
@@ -115,9 +115,10 @@ function libspell.GetSpellInfo(index, bookType)
 
   if id then
     scanner:SetSpell(id, bookType)
-    local _, sec = scanner:Find(gsub(SPELL_CAST_TIME_SEC, "%%.3g", "%(.+%)"))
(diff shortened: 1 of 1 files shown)


shagu:
Hi,
I don't understand how the added `false` would change anything here.
A non-existing param already is `nil` which has the same effect as `false` here.

What am I missing here?

dsidirop:
> Hi, I don't understand how the added `false` would change anything here. A non-existing param already is `nil` which has the same effect as `false` here.
> 
> What am I missing here?

That's in fact an excellent question - I should have made a better commit-comment to explain what's amiss. Take the following snippet for example:

```lua
local _, min = scanner:Find(gsub(SPELL_CAST_TIME_MIN, "%%.3g", "%(.+%)"))
```

One would think that this is equivalent to:

```lua
var x = gsub(SPELL_CAST_TIME_MIN, "%%.3g", "%(.+%)")
local _, min = scanner:Find(x)

shagu:
Thanks! A very good catch! :+1:

Issue #1290, opened 2024-06-02
Consolidate GetPlayerBuff() into a standard api method so that it's 1st parameter 'buffId' will be 1-based like TBC


dsidirop:
Current code:

```lua
local bid = GetPlayerBuff(PLAYER_BUFF_START_ID+id, type)
local stacks = GetPlayerBuffApplications(bid)
```

Suggested new approach:

```lua
local bid = pfUI.api.GetPlayerBuffX(id, type)
local stacks = GetPlayerBuffApplications(bid)
```

Note: It goes without saying that 'GetPlayerBuffX()' will NOT change the return value 'buffIndex' and we will leave it as-is for vanilla.

That's because if we also translated the returned value into a 1-based index (like in TBC) we would effectively break the rest of the Vanilla API surface that expects the returned 'buffIndex' to be 0-based in case of success and -1 in case of failure.

[ Unless ofcourse we also wrap all other Vanilla API methods like GetPlayerBuffTimeLeft() and so on ... which is an overkill ]

PS: In the context of this change I'll also make some trivial optimizations by consolidating the calls to GetPlayerBuffX() so that it will be called once when the exact same parameters are passed to it in adjacent rows in the code.

PR #1291, opened 2024-06-02
#1290 Consolidate getplayerbuff into an api method

dsidirop:
https://github.com/shagu/pfUI/issues/1290

### DIFF
--- a/api/api.lua
+++ b/api/api.lua
@@ -1482,11 +1482,13 @@ pfUI.api.PLAYER_BUFF_NOT_FOUND = pfUI.expansion == "vanilla" and -1 or 0
 -- Example of usage:
 --
 --     for i=1,32,1 do
---         local buffIndex, untilCancelled = GetPlayerBuffX(i, "HELPFUL")
+--         local buffIndex, untilCancelled = pfUI.api.GetPlayerBuffX(i, "HELPFUL")
 --         if buffIndex == pfUI.api.PLAYER_BUFF_NOT_FOUND then
 --             [...handle buff not found...]
 --         else
 --             [...handle buff found...]
+--
+--             local buffTexture = GetPlayerBuffTexture(buffIndex) -- this simply works properly on both vanilla in tbc
 --         end
 --     end
 --
--- a/api/api.lua
+++ b/api/api.lua
@@ -14,11 +14,15 @@ mod = math.mod or mod
 -- 'subject'    [string]        String to split.
 -- return:      [list]          a list of strings.
 function pfUI.api.strsplit(delimiter, subject)
-  if not subject then return nil end
-  local delimiter, fields = delimiter or ":", {}
-  local pattern = string.format("([^%s]+)", delimiter)
-  string.gsub(subject, pattern, function(c) fields[table.getn(fields)+1] = c end)
-  return unpack(fields)
+    if not subject then
+        return nil
+    end
+    local delimiter, fields = delimiter or ":", {}
+    local pattern = string.format("([^%s]+)", delimiter)
+    string.gsub(subject, pattern, function(c)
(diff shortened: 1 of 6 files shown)


shagu:
I always tried to move as much of vanilla-vs-tbc checks out of the code and use the compat ones instead.
But this is straight the other way around, moving even more vanilla-vs-tbc checks into the API code.

Also your IDE changed the formatting from 2 spaces to 4 spaces in the whole api.lua and also reformatted code blocks that have nothing to do with the PR itself.

I haven't checked the details yet, but if TBC already has a GetPlayerBuff() that we can use without modifications, it would be better to implement a version in compat/vanilla.lua that works the same. (Like done with hooksecurefunc and such)

dsidirop:
> But this is straight the other way around, moving even more vanilla-vs-tbc checks into the API code.
> I haven't checked the details yet, but if TBC already has a GetPlayerBuff() that we can use without modifications, it would be better to implement a version in compat/vanilla.lua that works the same.

Good points. I thought of doing that but bare in mind that there is a huge caveat that I already pointed out:

If I do override 'GetPlayerBuff' itself inside compat/vanilla.lua to make it behave like the tbc one then pfUI will work but any third party pfui-extension out there that uses 'GetPlayerBuff()' expecting to work as in vanilla will immediately break because they expect 0-based indeces but with this patch we'll be going behind their backs enforcing 1-based indeces on 'GetPlayerBuff()'.

Here's what I mean in code: 

```lua
setfenv(1, pfUI:GetEnvironment())  -- 3rd party addons import the pfUI environment

-- they naively use 0->31 because they aim only for vanilla stuff
-- this works for them but it will break once I patch GetPlayerBuff()!
-- it's very hard for them to figure out what broke and to change it to    for i=1,32,1   which is the tbc way of doing things!
for i=0,31,1 do 
       local b = GetPlayerBuff( i, "HELPFUL" )
       [...]
end
```

We can avoid all of this if I create under compat/vanilla.lua and compat/tbc.lua a side-flavour called 'GetPlayerBuffX()'. In TBC it will be set simply to 'GetPlayerBuff' but in vanilla it will be polyfilled in the same fashion as shown in this patch.

This way we won't cause any problems to the existing ecosystem that imports 'setfenv(1, pfUI:GetEnvironment())'.

Let me know your thoughts on this tradeoff.

shagu:
Well, that makes sense.. If I'd had to pick one, the 2nd approach seems to be the best solution, where we have PlayerBuffX in both compat/vanilla and compat/tbc.

I just wonder right now if it's even worth doing it: After all, it will be more lines of code, but with the same functionality as we already have. It just saves the user (mod developer) from placing a static variable into it, right?
A mod developer must still know the client difference otherwise he wouldn't even think of using GetPlayerBuffX instead of GetPlayerBuff. And at that point, he could also place the static variable in.

The next question we should ask ourselves is where do we stop? Are we going to build wrapper functions for every single constant variable in the compat module? To be honest, I have the fear to get lost in wrapper functions and the bloat around it.

What are your thoughts about that?

dsidirop:
>  To be honest, I have the fear to get lost in wrapper functions and the bloat around it.

I'm afraid of the same thing. This is why I frugal and only introduce the bare minimum of changes which insofar is just the first argument of GetPlayerBuffX() while keeping its return values the same as before.

Regarding the naming is 'GetPlayerBuffX()' a good naming convention? Or is something like (just thinking out loud) 'PFGetPlayerBuff()' better? (implying that future compat-utils will also be named as PF<something>)

Spend some time thinking about what naming convention rings true to you and let me know what you decided.

shagu:
Hey, I have decided to not merge this in. Thanks for your efforts tho.
But I do prefer the solution with static variables over the suggested one with wrapper functions.

Issue #1292, opened 2024-06-02
Optimize functions that use GetTime() multiple times by calling GetTime() just once (when possible/meaningful to do so)


dsidirop:
Given that GetTime() reaches out all the way to the underlying OS it kinda makes sense to snapshot and reuse the timestamps it generates when possible. An example of before-after:

Before:

```lua
visibilityscan:SetScript("OnUpdate", function()
  if (this.limit or 1) > GetTime() then return end

  this.limit = GetTime() + .2
  [...]
```

After:

```lua
visibilityscan:SetScript("OnUpdate", function()
  local now = GetTime()
  if (this.limit or 1) > now then
    return
  end

  this.limit = now + .2
  [...]
```

PR #1293, opened 2024-06-02
#1292 Optimize functions that use GetTime()

dsidirop:
https://github.com/shagu/pfUI/issues/1292

### DIFF
--- a/api/ui-widgets.lua
+++ b/api/ui-widgets.lua
@@ -1198,10 +1198,11 @@ function pfUI.api.CreateInfoBox(text, time, parent, height)
     infobox:Hide()
 
     infobox:SetScript("OnUpdate", function()
-      local time = infobox.lastshow + infobox.duration - GetTime()
+      local now = GetTime()
+      local time = infobox.lastshow + infobox.duration - now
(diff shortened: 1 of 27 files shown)


shagu:
Thanks for your suggestion, but I'll have to decline it.

I already thought about this myself and have decided to keep it as a single line.
Since every tick-line in pfUI basically looks the same, it's pretty easy to identify.

Even with caching it to local, most of the GetTime() will reach out anyways and only the GetTime() is saved
that is already limited inside the actual tick. I prefer to have it tidy in one line and to not put a new variable on the heap
in every iteration just to save one function call that is already throttled.

dsidirop:
> only the GetTime() is saved that is already limited inside the actual tick.

Keep in mind that the code as it is right now is there are quite a few cases where GetTime() is called repeatedly inside for-loops multiple times. It's beyond just aesthetics.

shagu:
Sorry, I have only checked the majority of this PR which have been the this.tick throttles.

You are right about the others. The ones that are repeatedly caleld inside for-loops like for debuffs and buffs, should really get cached only once per loop.

PR #1303, opened 2024-06-11
[libdebuff.lua] Align and optimize spell post-hooks

dsidirop:
The hooks can retrieve the name of the raw-effect from "libspell.GetSpellInfo()". In the case of "CastSpellByName" this will solve a bug which causes debuffs to not be detected optimally in case an explicit rank is passed to "CastSpellByName":

```
/script   CastSpellByName("<some spell>(Rank 2)")
```


dsidirop:
Any ideas on this? ( hope you're doing alright in rl btw )

shagu:
> Any ideas on this? ( hope you're doing alright in rl btw )

Sorry, I had to take a break.

PR #1304, opened 2024-07-03
[module] Bags - search pattern update

shikulja:
Just an attempt to make the search work in Russian.. and other languages Need testing.

image


shagu:
I'm not very good in regex, could you please describe in a few words what the new pattern does?
There's also this new issue open, which I think could be solved by this PR? (https://github.com/shagu/pfUI/issues/1324)

shagu:
Thanks! Can you please check the bag search with russian inputs on the latest version?

I just made a change simplifying the bag search: https://github.com/shagu/pfUI/commit/a9d2494fee1ce94b10099b766117260f5aa5e6d5

As far as I remember, the regex was only there to strip `%`, `[` and other patterns from the search, to prevent errors.
Basically what the 4th argument of string.find does, it disables pattern search as a whole.

shikulja:
works fine
image

special characters also do not cause errors
image

PR #1305, opened 2024-07-03
[module] Bags - update focus logic

shikulja:
Resident Evil 3 - Nemesis (Europe) 2024-07-04-01-45-08
Resident Evil 3 - Nemesis (Europe) 2024-07-04-01-43-36


shikulja:
Convenient when fishing, so that bags do not block the controls, when running in a raid with open bags..
In all cases where you need to actively move, or often focus on

Issue #1184, opened 2023-07-22
chat: copy mode [improvement]


shikulja:
Add: support for exiting copy chat mode using the esc key.
Or - deselect > esc > exit copy > esc

This is very lacking when you have to debug a new addon

PR #1306, opened 2024-07-04
[module] chatcopy - escape from chatcopy mode

shikulja:
https://github.com/shagu/pfUI/issues/1184


shagu:
Thanks, looks good so far. I will test and merge it next week

PR #1312, opened 2024-07-19
Add XY Offsets to Unit Power Bars

eirx32:
To allow more freedom with powerbar position:
image

Ideally it would be draggable or separatable from unit frame.
Or i could make a new powerbar frame that users can enable with a checkbox? Something like StandalonePowerbar?


shagu:
Hi, thanks for your contribution.
Wouldn't that, as it is now, break the combat glow sizing and the portrait sizes? I wont be able to access my PC the next days, but I will check out your branch and test it the next weeks

eirx32:
Hey, thanks for responding.
I just realized i messed up users on GitHub, so i will close this PR and make a new one when i fix them, very sorry for the inconvenience.

PR #1317, opened 2024-08-01
Add Nameplate Height Offset

mrrosh:
image


shagu:
It doesn't change the clickable area, resulting in an unclickable nameplate with an invisible clickable square below.
I've been there and done that, resulting in too many requests and complains about how buggy it is. I will not do it again.

PR #1319, opened 2024-08-18
[Loot] Fix infinite loop autopickup. replace with delayed ConfirmLootSlot()

AiAnswer3213:
Could not avoid the delay.
Instead of clicking accept and risk clicking wrong popup, just use the proper confirm and close the dialog

tested only 2.4.3

### DIFF
--- a/modules/loot.lua
+++ b/modules/loot.lua
@@ -715,7 +715,7 @@ pfUI:RegisterModule("loot", "vanilla:tbc", function ()
         local slot = arg1
         QueueFunction(function()
           ConfirmLootSlot(slot)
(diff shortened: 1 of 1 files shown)


ksidirop-laerdal:
> Could reproduce it working whilst picking up some [Demonic Runes](https://www.wowhead.com/classic/item=12662/demonic-rune)

Did you test in in vanilla wow 1.12.x or in a TBC 2.x client?

shagu:
Sadly it doesn't work on vanilla:
image

I have adapted your changes and pushed it via: https://github.com/shagu/pfUI/commit/21f77eddbdded806681f435bec85cc048156d5ed

Thank you very much for your contribution! :)

PR #1322, opened 2024-09-04
unitframes: add left/center/right text offset for hp & power bars.

dbulatovicx32:
Added default values, logic and gui for text offsets, few things to note which might be good reasons not to accept the PR in its current state:

**1.** I did not change anchor points so existing UI wouldn't break, however using extreme X offset values with **TOPLEFT** and **BOTTOMRIGHT** anchors may cause text to disappear.

_eg: right text 20 X offset applied, but exits its anchor so its truncated_
image

**2.** **center** text cannot apply its X offset **at all** because of anchors mentioned above. 

_eg: center text -5 X offset value not applied_
image

**3.** I'm worried that GUI might become too cluttered with new input fields, possible solution is to separate them somewhere else, or hide offset inputs if text is disabled/hidden.

Overall it works as a proof of concept, but I'm not happy with its current state, id like to change the anchors in a way that wont break the original UI, but didn't want to touch it without your permission first ~_~


Kevin-Yxm:
Thank you very much. Very useful. Adding a font size adjustment would be perfect

shagu:
Sorry for the huge delay & Thanks for your work. It's now merged :)

shagu:
... And fixed: https://github.com/shagu/pfUI/commit/9bb9dbf4e1149de964e1f19b4851f0040ae94e7e

PR #1326, opened 2024-09-22
Added option to anchor bags above chat

bhhandley:
...Actual PR ... somehow undid the changes for bank bags before committing.

Always preferred bags being above chat so I can still read and interact with it while going through my bank/inventory.  This adds a checkbox to the Bags & Bank settings so you can choose to anchor your bags to the top or bottom of the chat windows.


shagu:
Looks good! Thank you. Merged :)

PR #1327, opened 2024-09-24
Added total line to gold info panel.  Added (rough) translations.

bhhandley:
The information layout on the gold tooltip makes it hard to quickly add up your total gold across characters with a quick glance.  Added a total line so I didn't have to do the math.  Most of the German translations were set to nil so I added the same, the others are rough machine translations.


shagu:
Thanks! And sorry for the delay.

PR #1337, opened 2024-10-30
rearrange twow's profession window additions

Zebouski:


shagu:
7h4nk5, l00k5 600d 70 m3!

PR #1343, opened 2024-11-03
turtle-wow: update Immolate duration after cast Conflagrate

Artur91425:
In 1.17.2 TW devs change Conflagrate. Now it consumes 3 sec Immolate.
You can read new talent description here:
https://talents.turtle-wow.org/warlock


shagu:
Thanks! :+1:

PR #1345, opened 2024-11-04
castbar: set strata to high so it can overlap unitframes

Pizzahawaiii:
Increase castbar frame strata so it's able to overlap unitframes again.

I positioned the castbar on top of my unitframes to simulate it replacing the powerbar. That was broken in commit [afba5f](https://github.com/shagu/pfUI/commit/afba5fe60bed6dab5095a5272e1328f829512a73) and this PR fixes it again.

pfui-castbar-fix

### DIFF
--- a/modules/castbar.lua
+++ b/modules/castbar.lua
@@ -8,7 +8,7 @@ pfUI:RegisterModule("castbar", "vanilla:tbc", function ()
     local cb = CreateFrame("Frame", name, parent or UIParent)
 
     cb:SetHeight(C.global.font_size * 1.5)
-    cb:SetFrameStrata("MEDIUM")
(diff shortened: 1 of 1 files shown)


shagu:
Oh god i had frame FrameStratas... Thank you.
Have you checked other frames after that change?

shagu:
With the suggested change, the castbar would be above the Spellbook, Character Frame, Bags, etc.

To stop the cascade of always increasing FrameStratas, I instead changed the FrameLevel only via: https://github.com/shagu/pfUI/commit/f420c5f7f7e111c66c98abc8a1e3bf0ed658ed28

Thanks for your report and contribution :)

PR #1360, opened 2024-11-08
buff: add option to align buffs left so they grow to the right

Pizzahawaiii:
Add option to left-align buffs/debuffs/wepbuffs. Seems to work fine both with and without separate weapon buffs.

Fixed it for weapon buffs by making weapon buff frame as wide as debuff frame. Feels a bit hacky but I don't see how that could cause problems. Do you? :see_no_evil: 

Screenshot from 2024-11-08 23-22-45

Screenshot from 2024-11-08 23-22-01

Screenshot from 2024-11-08 23-22-09

Screenshot from 2024-11-08 23-22-18

Screenshot from 2024-11-08 23-22-25


shagu:
If I remember correctly, we have agreed via discord to leave this unmerged because of too many things that broke during testing, right? I'm not sure if you already had time and motivation to fix those. 

Please feel free to reopen this again when it's in a better shape. Thank you :)

PR #1362, opened 2024-11-09
skin: QuickHealHealingBar

shikulja:
Maybe later I'll add a QH menu hook


shikulja:
Glassic style
image

shikulja:
for colouring, can be used, pfui globals but i not sure
only found
C.chat.global.background
C.chat.global.border

shagu:
I really like to move the skins from thirdparty-vanilla.lua to the skin directory in future.
It makes sense to have them there. Just saying what you're doing is the right thing, but I can't merge it until that transition is done. Skins should be either in thirdparty-vanilla.lua or in skins/, so I will leave this PR open until then.

Please continue what you're doing, happy holidays and thank you for your contributions :)

PR #1380, opened 2024-12-06
Fixed pet happiness & totems for Totemic Recall

jmgibson1981:
Probably a better way to handle the totems. Pet happiness all I did was disable for warlock.

### DIFF
--- a/api/unitframes.lua
+++ b/api/unitframes.lua
@@ -1411,7 +1411,7 @@ function pfUI.uf:RefreshIndicators(unit)
   end
 
   if unit.happinessIcon and unit:GetName() == "pfPet" then -- Happiness Icon
-    if unit.config.happinessicon == "0" then
+    if unit.config.happinessicon == "0" or UnitClass("player") == "Warlock" then
       unit.happinessIcon:Hide()
     else
       if UnitIsVisible("pet") then
--- a/libs/libtotem.lua
+++ b/libs/libtotem.lua
@@ -83,8 +83,12 @@ local totems = {
   },
 }
 
-GetTotemInfo = function(id)
+GetTotemInfo = function(id, totrecall)
   if not active[id] or not active[id].name then return end
+  if totrecall then
+    libtotem:Clean(id)
+    return nil
+  end
   if active[id].start + active[id].duration - GetTime() < 0 then
     libtotem:Clean(id)
     return nil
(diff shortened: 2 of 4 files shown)


shagu:
Thanks for your contribution, but I can't merge it like that.
It has way to many locale specific code and turtle things mixed into the main modules.
Also the additional changes to the saved variables is not what I want to have like that.

jmgibson1981:
No worries. Thank you for responding. Im very new to Lua. As I stay with
turtle I’ll keep it forked but I’ll look for other ways i can help. Thank
you again.

shagu:
> Im very new to Lua.

Oh, okay. Here are some thoughts: I don't have a shaman to test it, but I'd suggest to workaround any turtle-wow specific changes from within the turtle-wow module. For the totems, I think it could be possible to create a frame that listens on the `CHAT_MSG_SPELL_SELF_BUFF` to check for the given string, and then run: `for i = 1,4 do libtotem:Clean(i) end`

So it would look like something like this:
```lua
local recall = CreateFrame("Frame")
recall:RegisterEvent("CHAT_MSG_SPELL_SELF_BUFF")
recall:SetScript("OnEvent", function()
  if arg1 and string.find(arg1, "You gain") and string.find(arg1, "Mana from Totemic Recall") then
    for i = 1,4 do 
      libtotem:Clean(i) 
    end
  end
end)
```

This way, we wouldn't need to touch libtotem for turtle changes at all, and could just remove all totems from within the turtle-wow module when their custom spell is detected.
Next would be to look into the regex, the goal would be to build it with just one string, that can then be localized easier. Like changing:
```lua
if arg1 and string.find(arg1, "You gain") and string.find(arg1, "Mana from Totemic Recall") then
```

to something like this:
```lua
if arg1 and string.find(arg1, T["You gain (.+) Mana from Totemic Recall"]) then
```

With just one string and the surrounding of `T[]` it can be later easilly translated, to make it work on a zhCN client aswell. This is all untested and I just wrote it of the top of my head, but that would be a cleaner approach to the totem issue that will for sure get merged.

Regarding the warlock-pet happiness, the problem is UnitClass()'s first return value is a localized one. So it would need to be done different in all languages. I suggest to use the second return value of it, which is an uppercase english name.
```lua
local _, myclass = UnitClass("player")
```
You would then check for `myclass == "HUNTER"` which then works on all game languages :)

Feel free to send two individual pull requests for both issues, and we will get it merged :+1: 
If you don't want to, that's fine too. I will maybe pick it up one day and do it myself.

Also feel free to ask whenever in doubt of anything lua related, I'm happy to help and explain.

Cheers
Shagu

PR #1387, opened 2024-12-27
Added/changed options for minimap zone and coordiates visability.  Added/updated translations

bhhandley:
- Added a dropdown for coordinates to be either on mouseover, off, or on
- Moved the existing "off" option from the locations into the new dropdown
- Changed the zone text option from a checkbox to a dropdown with the options of mouseover, off or on
- Moved most of the OnEnter events to an OnEvent tied to MINIMAP_ZONE_CHANGED and PLAYER_ENTERING_WORLD
- Changed the default coordinate location to bottomleft so it doesn't overlap the zone text by default
- Added/updated translations for the options.

image


shagu:
Hello, thanks for your contribution. I really like the idea and the ability to configure it like that.

But sadly, I can't merge it as is, because it lacks the default configs in api/config.lua and it breaks existing configs. You removed entries from the dropdown lists and changed the behaviour of the settings without adding the required config migrations for it.

pfUI always had the goal to keep the config stable. So that people won't lose their settings or have weird looking setups after performing an update. There's a whole section in api/config.lua that only covers migrations over breaking config changes.

bhhandley:
Updated, sorry for missing that.

shagu:
Thanks for you quick update. There still seems to be an issue for me: Setting both `Zone Text On Minimap` and `Coordinates On Minimap` to `Enable` will reset `Coordinates on Minimap` back to `Mouseover` after reload.

Haven't looked into it in detail yet. Maybe you already know.

bhhandley:
Updated.  Yep, I was only testing the initial load to make sure the old settings migrated.  Forgot to have the migration test to make sure the new settings don't exist.

shagu:
The settings seem to be swapped
https://github.com/user-attachments/assets/f869e1ea-5a47-44fb-982e-2e5456144f80

shagu:
Looks fine to my after applying:
```diff
diff --git a/modules/minimap.lua b/modules/minimap.lua
index eea574c9..712b46a3 100644
--- a/modules/minimap.lua
+++ b/modules/minimap.lua
@@ -207,10 +207,10 @@ pfUI:RegisterModule("minimap", "vanilla:tbc", function ()
   end)
   Minimap:SetScript("OnLeave", function()
     if C.appearance.minimap.zonetext ~= "on" then
-      pfUI.minimapCoordinates:Hide()
+      pfUI.minimapZone:Hide()
     end
     if C.appearance.minimap.coordstext ~= "on" then
-      pfUI.minimapZone:Hide()
+      pfUI.minimapCoordinates:Hide()
     end
   end)
```

shagu:
Thanks for you contribution. Looks good to me now :+1:

bhhandley:
... wow.  Well, thanks for catching and fixing it.

shagu:
Looks like there was another one :see_no_evil: https://github.com/shagu/pfUI/commit/cedee86c4cca5ec3e2ea2791e1d3d351cecfa60c

PR #1390, opened 2024-12-30
map: ensure strata is fullscreen

gashole:
This fixes a minor issue encountered on the Turtle WoW server, where switching from a windowed map to using pfUI caused the strata to be set to medium until the UI was reloaded.

### DIFF
--- a/modules/map.lua
+++ b/modules/map.lua
@@ -39,6 +39,7 @@ pfUI:RegisterModule("map", "vanilla:tbc", function ()
 
     UIPanelWindows["WorldMapFrame"] = { area = "center" }
 
+    WorldMapFrame:SetFrameStrata("FULLSCREEN")
     WorldMapFrame:SetMovable(true)
     WorldMapFrame:EnableMouse(true)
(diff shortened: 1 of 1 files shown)


shagu:
Hey, could you please elaborate more? With pfUI the turtle feature of fullscreen/window map is disabled.
Where exactly does this fix something? I would like to see the issue before merging things.

gashole:
Sure, the strata is expected to be fullscreen, but pfUI doesn’t explicitly set it. This commit addresses that. When Turtle's windowed map is enabled before pfUI, it lowers the strata. Since pfUI doesn’t update it, the strata remains lowered until the UI is reloaded. The fix is simple: set the strata to fullscreen. Here’s a clip showing the issue: https://youtu.be/l2oNhBGnxsA

shagu:
That's interesting. But it seems to be a very corner case issue and also a turtle-wow specific one.
I'd prefer to not mess around with FrameStratas unless really required, especially if it's such minor and custom client change related. But thanks you for letting me know and taking your time for this anyways! :+1:

PR #1394, opened 2025-01-04
adjusted for totemic recall

jmgibson1981:
your code worked flawlessly. much simpler and i now understand modules a bit better. thank you. I've no idea how to test localization.

### DIFF
--- a/modules/totems.lua
+++ b/modules/totems.lua
@@ -15,6 +15,16 @@ pfUI:RegisterModule("totems", "vanilla:tbc", function ()
     totems:RefreshList()
   end)
 
+  local trecall = CreateFrame("Frame", "pfTotemsRecall", UIParent)
+  trecall:RegisterEvent("CHAT_MSG_SPELL_SELF_BUFF")
+  trecall:SetScript("OnEvent", function()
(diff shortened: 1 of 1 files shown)


shagu:
Nice! I'm happy it was helpful. You don't need to care or test anything about translations.
As soon as a string is defined as `T["something"]`, translators can pick it up if they wish to.

However, Totemic Recall is a turtle wow custom spell that does not exists in vanilla.
As mentioned on the other PR, I'd prefer to have all turtle wow  custom things in one place.
So it would be nice if you could move it to the [turtle-wow module](https://github.com/shagu/pfUI/blob/master/modules/turtle-wow.lua#L13).

You can just put your stuff there, with a short comment explaining what it does, like:

```lua
  -- add support for turtle wow's totemic recall spell
  local _, class = UnitClass("player")
  if class == "SHAMAN" then
    ...
  end
```

Just have a quick look at the other turtle-wow fixes and you should understand. :)
(It's also a nice place to learn how to modify pfUI internals from within another module)

Thanks for your work! :+1:

jmgibson1981:
Ok ill look into the specific modules.

PR #1402, opened 2025-01-22
soulshards count stackable

jmgibson1981:
Switched soul shards to count by stack as Turtle, and I assume some others may have them stackable.


shagu:
Thanks!

Issue #1403, opened 2025-01-23
Pet Autocast stays visible with bar on autohide


jmgibson1981:
On initial pet summon anything that has autocast enabled, the autocast stays enabled despite the bar being set to autohide.

https://i.ibb.co/XLbQSpC/Screenshot-20250122-224021.png

They fade as they should once I've moused over the bar then left it again but that first bar "creation" fails to keep the autocast invisible.

I've tried digging around in it but it's proving elusive. my lack of lua skill isn't helping.

shagu:
To get this done, i want to reproduce it on my end first and also look into whats happening.
Sadly I don't have much time recently. It will be delayed for a little more.

shagu:
(fixed via: https://github.com/shagu/pfUI/commit/901b482de2eee7956a4975cfd725754015526810)

PR #1406, opened 2025-01-29
fixed autocast show while bar autohide active on create

jmgibson1981:
This solves [issue 1403](https://github.com/shagu/pfUI/issues/1403). The only issue is that until a button update happens the alpha of the autocast texture is 100%. Once autocast has been toggled or the bar does an update it all works correctly. I'm unsure of the best way to approach this, if it's even worth fixing the temporary 100% alpha on it.

I did try calling a button update but it had no effect.

*EDIT* Upon further testing it appears this 100% is the default behavior anyway so I won't chase it any further.


shagu:
Sorry, don't have much time recently. I'll postpone looking into this a bit further

PR #1415, opened 2025-02-06
Panel Additions For Rogue

jmgibson1981:
I added a couple panel indicators. Flash Powder and Thistle Tea. Also an indicator showing current bind / hearth location.

### DIFF
--- a/modules/gui.lua
+++ b/modules/gui.lua
@@ -1069,7 +1069,10 @@ pfUI:RegisterModule("gui", "vanilla:tbc", function ()
         "combat:" .. T["Combat Timer"],
         "ammo:" .. T["Ammo Counter"],
         "soulshard:" .. T["Soulshard Counter"],
-        "bagspace:" .. T["Bagspace"]
+        "bagspace:" .. T["Bagspace"],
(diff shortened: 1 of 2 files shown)


shagu:
There's already a function to find an item count by name:
https://github.com/shagu/pfUI/blob/master/api/api.lua#L294-L317

Unless there are reasons behind your solution, i would suggest to use the already existing one.

jmgibson1981:
No problem. I didn't realize that. I moved the one out of the
Soulshards and made it global to the file. I'll look for the existing
one.

On Mon, 2025-02-10 at 01:25 -0800, Shagu wrote:
> There's already a function to find an item count by name:
> https://github.com/shagu/pfUI/blob/master/api/api.lua#L294-L317
> Unless there are reasons behind your solution, i would suggest to use
> the already existing one.
> —
> Reply to this email directly, view it on GitHub [1], or unsubscribe
> [2].
> You are receiving this because you authored the thread.Message ID:
> ***@***.***>


[1] view it on GitHub
    https://github.com/shagu/pfUI/pull/1415#issuecomment-2647406702
[2] unsubscribe

PR #1418, opened 2025-02-10
Fixed up code reduction

jmgibson1981:
Also changed warlock soulshards to use the item  count in the api
Added a Hearthstone, Thistle Tea, and Flash Powder panel items.


shagu:
I have reverted this PR again. It turns out it had too many flaws and I don't have the time to work on it right now.

jmgibson1981:
Damn. I don’t move mine so never noticed. I’ll look into it. Apologies.

PR #1419, opened 2025-02-10
Overwrite nameplate color and border with agro state

Meanmage:
Inspired by TankPlates by MarcelineVQ which is a great addon but makes you lag a lot. Therefore this implementation into pfUI which avoids the lag penalty.

Equivalent game logic from TankPlates has been implemented except no recoloring of nameplates to orange if the unit is affected by a specific debuff from a list of crowd controls.

Further considerations: 
- How it should interact with the "Overwrite Border Color With Combat State". Currently this new feature overwrites it. Maybe make it a drop down menu for exclusive choice?
- Removing in combat restrictions for better PvP usefulness.

### DIFF
--- a/api/config.lua
+++ b/api/config.lua
@@ -725,6 +725,7 @@ function pfUI:LoadConfig()
   pfUI:UpdateConfig("nameplates", nil,           "totemicons",       "0")
   pfUI:UpdateConfig("nameplates", nil,           "showguildname",    "0")
 
+  pfUI:UpdateConfig("nameplates", nil,           "agrostate",        "0")
   pfUI:UpdateConfig("nameplates", nil,           "outcombatstate",   "1")
   pfUI:UpdateConfig("nameplates", nil,           "outfriendly",      "0")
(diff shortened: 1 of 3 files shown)


Flecked42:
WoWScrnShot_021525_151735 tga

The mob names turning orange, is that normal?

Meanmage:
> Very cool, much better then the border outline.
> 
> WoWScrnShot_021525_151735 tga
> 
> The mob names turning orange. I'm not sure what that means, is that normal?

Yes it's normal behavior whenever you damage a mob. It's the same on production.

Flecked42:
Could this be disabled? More of a question for Shagu.

Also Its difficult to tell which mod you have targeted sometimes.  Maybe the target name could be changed yellow like in the tank plates addon?

Meanmage:
There is the "Draw Glow Around Target Nameplate" and "Zoom Target Nameplate" for that

Flecked42:
I use the glow, but its really hard to notice. I'll have to try out the zoom option.

PR #1427, opened 2025-02-17
panel: add new options and reduce code

jmgibson1981:
UpdateMovable was the culprint. Moved it back out roughly to it's original location. Panels stay where you put them.


modules/panel.lua
```diff
@@ -621,10 +656,62 @@ pfUI:RegisterModule("panel", "vanilla:tbc", function()
...
+      frame.text:SetAllPoints(frame)
+      frame.text:SetPoint(location, 0, 0)
+      frame.text:SetJustifyH(tjustify)
+      frame.text:SetFont(font, font_size, "OUTLINE")
+      frame.text:SetFontObject(GameFontWhite)
+      return frame
+  end
+
+  local function CreatePanel(panelname, default_border)
+    frame = CreateFrame("Frame", panelname, UIParent)
```
shagu, line 677: I think this should be `local`

modules/panel.lua
```diff
@@ -621,10 +656,62 @@ pfUI:RegisterModule("panel", "vanilla:tbc", function()
     end
   end
 
+  local function CreatePanelButton(parent, width, location, tjustify)
+    local frame = CreateFrame("Button", nil, parent)
+      frame:SetFrameLevel(0)
```
shagu, line 661: Looks like the indentation is wrong from here on. Should be indented by 2 spaces, not 4.

modules/panel.lua
```diff
@@ -554,25 +554,60 @@ pfUI:RegisterModule("panel", "vanilla:tbc", function()
...
+      end)
+    end
+
+    do -- Hearthstone bind location
+      local widget = CreateFrame("Frame", "pfPanelBindLocation", UIParent)
+      widget:RegisterEvent("PLAYER_ENTERING_WORLD")
+      widget:RegisterEvent("CHAT_MSG_SYSTEM")
+      widget:SetScript("OnEvent", function()
+        local bindlocation = GetBindLocation()
+        pfUI.panel:OutputPanel("bindlocation", T["Hearthstone"] .. ": " .. bindlocation)
```
shagu, line 571: Have you checked this doesn't throw an `trying to concatinate a nil value`-error if no bind location is set?

jmgibson1981: Just pushed an update for a nil check and an output of Not Set if no bind.

modules/panel.lua
jmgibson1981: fixed that.

modules/panel.lua
jmgibson1981: totally missed it. ok all 3 fixed  ready to roll.

modules/panel.lua
```diff
@@ -554,25 +554,65 @@ pfUI:RegisterModule("panel", "vanilla:tbc", function()
...
+      widget:RegisterEvent("PLAYER_ENTERING_WORLD")
+      widget:RegisterEvent("CHAT_MSG_SYSTEM")
+      widget:SetScript("OnEvent", function()
+        local bindlocation = nil
+        bindlocation = GetBindLocation()
+        if bindlocation ~= nil then
+          pfUI.panel:OutputPanel("bindlocation", T["Hearthstone"] .. ": " .. bindlocation)
+        else
+          pfUI.panel:OutputPanel("bindlocation", T["Hearthstone"] .. ": " .. T["Not Set"])
+        end
```
shagu, line 576: As this PR is mainly about code reduction, that whole part could also be reduced this:
```lua
pfUI.panel:OutputPanel("bindlocation", T["Hearthstone"] .. ": " .. (GetBindLocation() or T["Not Set"]))
```

:)

PR #1428, opened 2025-02-17
nameplates: overwrite healthbar and border color with agro state

Meanmage:
Inspired by TankPlates by MarcelineVQ which is a great addon but makes you lag a lot. Therefore this implementation into pfUI which avoids the lag penalty.

Equivalent game logic from TankPlates has been implemented except no recoloring of nameplates to orange if the unit is affected by a specific debuff from a list of crowd controls.

Further considerations:
- Adding the logic of recoloring the nameplates orange if the unit is affected by hard CC.
- Removing in combat restrictions for better PvP usefulness.

### DIFF
--- a/api/config.lua
+++ b/api/config.lua
@@ -728,6 +728,7 @@ function pfUI:LoadConfig()
   pfUI:UpdateConfig("nameplates", nil,           "totemicons",       "0")
   pfUI:UpdateConfig("nameplates", nil,           "showguildname",    "0")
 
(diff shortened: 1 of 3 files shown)


shagu:
I have now added the same functionality for the healthbar colors that already existed for the healthbar outline. There's also an option now, for each of the 3 colors to customize them to your likings. Stuff like the castbar-logic is still missing.
Anyways, I'm closing this PR, please create a new one if you'd like to propose new changes. Thank you!

PR #1432, opened 2025-03-01
Rogue poisons panel widget

jmgibson1981:
Added a mouseover widget to display number of poisons in inventory.


shagu:
Thanks for your PR, but these are getting too many items now. I think another solution must be found. Maybe where people can just type the item they want to search for manually. Or a list of items, like in the buff-tracker whitelists.

I also just noticed the item names are not localized in your last PR. I will add localization to it later, but I'd like to avoid having a list of all existing items someday due to this. I think letting the user type in the name would be the better approach.

PR #1462, opened 2025-04-19
targetarg support for superwow/cleveroidmacros

jrc13245:
In order to allow pfui to support @mouseover casted buff and debuff timers with superwow/cleveroidmacros, i have edited the libpredict.lua. 
When CleveRoids.hasSuperwow is true, CleveRoids calls CastSpellByName with two arguments. The hook sees targetArg is not nil and uses the target provided by CleveRoids.
When CleveRoids.hasSuperwow is false (or for any other addon/default UI call), CastSpellByName is called with only one argument. The hook sees targetArg is nil and uses libpredict's original logic (checking pfUI mouseover or the current target).

### DIFF
--- a/libs/libpredict.lua
+++ b/libs/libpredict.lua
@@ -399,18 +399,33 @@ hooksecurefunc("CastSpell", function(id, bookType)
   spell_queue[3] = UnitName("target") and UnitCanAssist("player", "target") and UnitName("target") or UnitName("player")
 end, true)
 
-hooksecurefunc("CastSpellByName", function(effect, target)
-  if not libpredict.sender.enabled then return end
-  local effect, rank = libspell.GetSpellInfo(effect)
-  if not effect then return end
-  local mouseover = pfUI and pfUI.uf and pfUI.uf.mouseover and pfUI.uf.mouseover.unit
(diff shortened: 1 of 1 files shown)


shagu:
Hello, I don't understand why you added that many changes to the PR.
You renamed variables and removed the check to disable libpredict when healcomm is detected.

As far as I can tell, this seems to be the only required change:

```diff
diff --git a/libs/libpredict.lua b/libs/libpredict.lua
index d288253a..3a96f293 100644
--- a/libs/libpredict.lua
+++ b/libs/libpredict.lua
@@ -410,7 +410,7 @@ hooksecurefunc("CastSpellByName", function(effect, target)
 
   spell_queue[1] = effect
   spell_queue[2] = effect.. ( rank or "" )
-  spell_queue[3] = mouseover or default
+  spell_queue[3] = target and UnitName(target) or mouseover or default
 end, true)
 
 local scanner = libtipscan:GetScanner("prediction")
```

I did some tests with it and it works fine.
Thanks for your PR, but I like to have minimal changes and will do it my way. :)

shagu:
Added via: https://github.com/shagu/pfUI/commit/bedd293ce73cb5776a3d0f8981c5573303ce6c0e

PR #1463, opened 2025-04-20
prevent errors in libpredict from invalid passed targets

jrc13245:
sorry for the silly previous pull request, im not very good at lua. this should prevent invalid target errors, such as numbers. my original error came from using pfui autostance + cleveroidmacros attempting to cast spells that are not available in that stance.


shagu:
Fixed via: https://github.com/shagu/pfUI/commit/af2278da6bbb8cc1034ce5b178f2ce4df7a18f58

PR #1469, opened 2025-04-24
fixes earthshaker slam in tracking display on turtle

jmgibson1981:
I'm unsure if there is a better way as in comments. Earthshaker Slam shaman ability shows up in tracking icon for Turtle WoW. This skips it when loading known tracking modes.

### DIFF
--- a/modules/tracking.lua
+++ b/modules/tracking.lua
@@ -41,6 +41,10 @@ pfUI:RegisterModule("tracking", "vanilla", function ()
     }
   }
 
+  local invalidTrackingSpells = {
+    ["Earthshaker Slam"] = true
+  }
+
   local state = {
     texture = nil,
     spells = {}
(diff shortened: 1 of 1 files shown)


shagu:
Hello,
I like the solution with the invalid spells much more. But I'd prefer to not put Turtle-WoW related code into the main modules. I suggest to make the `invalidTrackingSpells` globally accessible, like via `pfUI.tracking.invalidTrackingSpells = {}` and then fill the values from within the `turtle-wow.lua` module. 

The function you have created seems to be useless and recreated on each iteration. You could remove it and just use the table index check instead: `if not pfUI.tracking.invalidTrackingSpells[validTracking] then`

shagu:
Had to restructure the tracking module a bit: https://github.com/shagu/pfUI/commit/a5103912b48f9efa6585d1c6bd37b4d09a398438 https://github.com/shagu/pfUI/commit/b0141ddf27cd85d212888ca025d3c9dc4cbc771f

Anyways, `Earthshatter Slam` is now blacklisted from the tracking menu via: https://github.com/shagu/pfUI/commit/39db30cce3203aa58a1f9acd13c6a05509e45116

PR #1478, opened 2025-05-07
unitframes: fix faulty powerbar center text x offset

dbulatovicx32:
Because center text is connected to powerbar, f.config.poffx is no longer needed to calculate the offset. This fixes an oversight of my previous PR.

In the images attached, notice the offset on center text even though value of offset is set to 0.

Before:
image

After: 
image


shagu:
Thank you :+1:

PR #1479, opened 2025-05-08
castbar: add XY text offset

dbulatovicx32:
- added XY offset to left/right castbar texts for player, target and focus
- reordered GUI so its consistent with unitframe config, width/height and disable on top

discord: eirx32

Feature Example:
image


shagu:
Hello, thanks for your PR!
Please fix the formatting offsets in the api/config.lua file and this will be ready to merge! :fireworks:

dbulatovicx32:
Done o7

PR #1480, opened 2025-05-08
cooldown: custom font

dbulatovicx32:
- added an option to freely change cooldown font, instead of it being hardcoded to font_unit

Potentially destructive because everyone's font is going to be reset to BigNoodle which is the default.
If you think this can be a problem, I could add a override font checkbox and create an IF condition instead of replacing old font.

discord: eirx32

Feature Example:
image


shagu:
I like this. There are no additional checkboxes or if statements required.
You can add a config migration that sets the Cooldown Text Font to the current unit font, and I will bump the version afterwards. This way, the migration should only apply once to those that upgrade, and will not change afterwards anymore. The migration stuff can be found in api/config.lua and is a one-way config change, that will only run if the current config version is lower than the current addon version. Once it is run, it sets the current config version to the current addon version to not run again.

dbulatovicx32:
Assuming something like that?
Its interesting that LoadConfig() already had set option for appearance.cd.font even thought GUI didn't exist for it.

dbulatovicx32:
Updated as requested and merged with master branch o7
Thanks for clarifying the oversight, it makes sense >.<

PR #1481, opened 2025-05-09
actionbar: custom icon size ratio

dbulatovicx32:
- added option to individually set height and width of an icon, rather than just setting the size for both
- modified BarLayoutSize to allow new values, probably should make my own helper for this?

One thing that might be worth looking into is how to keep the icon image ratio itself, rather than letting it squish like in the screenshots, it doesn't bother me personally, but it might bother you.

Default, using only Icon Size of 50:
image

Custom Size, using width:50, height:30:
image

Drag background is correctly scaled as well:
image


shagu:
This looks fine for me. The stretching doesn't bother me personally.
However, I'd like to verify that you have tested every corner case of it. There just have been too many unforseen `BarLayoutSize` issues from recent commits. Would be great if you could confirm it is working on `vertical` and `horizontal` bars, aswell as on bars with different sizes.

One more note, for me it would also be fine to move away from Icon Size at all, and instead use width/height values by default. That would save 2 config entries and the migration of it would be rather simple. What do you think?

dbulatovicx32:
> There just have been too many unforseen `BarLayoutSize` issues from recent commits. Would be great if you could confirm it is working on `vertical` and `horizontal` bars, aswell as on bars with different sizes.
>
Il sit down after work and thoroughly test them, i completely forgot to test horizontal bars and pet/totem etc.

> One more note, for me it would also be fine to move away from Icon Size at all, and instead use width/height values by default. That would save 2 config entries and the migration of it would be rather simple. What do you think?

I do agree, however afaik Bartender doesnt have this option and elvUI: retail didnt have it for a while, but they added it as a toggle:
_in elvui, you toggle off the checkbox and it displays height/width settings_
image

I would still leave the good old Icon Size because i think 90% of users wont use the custom ratio and they will have no issues with the feature in case something is wrong with `BarLayoutSize` unless they opt in to use it.
But this opinion also comes from my fear of not knowing the intricacy of `actionbars` module and how others interact with it, eg: `cooldowns` something i need to test as well.

PR #1496, opened 2025-05-25
fix: priest hot renew duration 15 sec

seybsen:
also there is no 5-piece bonus for renew


seybsen:
> > also there is no 5-piece bonus for renew
> 
> https://www.wowhead.com/classic/item=21349/footwraps-of-the-oracle

sorry, I forgot about that. I'm playing twow where this set is diffrent 😉 

https://github.com/shagu/pfUI/blob/7b4c839dbddd1fe3acfbc94421d1c6e2becac09a/libs/libpredict.lua#L360
needs to be adapted as well then:

```
renewDuration = setBonusCounter == 5 and 18 or 15
```

PR #1511, opened 2025-06-07
fix: handle 24-hour to 12-hour clock conversion edge cases. 00:00 and…

bennebbenneb:
# Overview
There is a bug when converting the 24-hour clock to a 12-hour clock.

## Issue
00:00 -> 00:00am
12:00 -> 12:00am

## Correct way
00:00 -> 12:00am
12:00 -> 12:00pm

These updates should fix the issue


shagu:
Thanks for your PR :+1:

Issue #1513, opened 2025-06-08
[trivial cleanups] chat.lua: dynamically generate pfUI.chat.URLFuncs() and have pfUI.chat:HandleLink() iterate through the pfUI.chat.URLPattern table


dsidirop:
Essentially what the subject says. If I understand correctly the order in which the patterns (WWW, PROTOCOL, ...) are iterated is not really that important (correct me if I'm missing something obvious here!)

Will create a quick PR about this.

shagu:
The idea was to be able to disable some of the patterns, which I never got to. 
It also helped in the past while debugging to see which of the patterns was used.

PR #1514, opened 2025-06-08
#1513 [trivial cleanups] chat.lua: dynamically generate pfUI.chat.URLFuncs() and have pfUI.chat:HandleLink() iterate through the pfUI.chat.URLPattern table

dsidirop:
https://github.com/shagu/pfUI/issues/1513

### DIFF
--- a/modules/chat.lua
+++ b/modules/chat.lua
@@ -127,16 +127,16 @@ pfUI:RegisterModule("chat", "vanilla:tbc", function ()
       ["fm"]="%s.%s.%s"},
   }
 
-  pfUI.chat.URLFuncs = {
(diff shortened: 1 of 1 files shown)


shagu:
I feel like the inline function call makes it less readable. And it doesn't really save much lines anyway.

I'm usually all for saving lines of code. But if it's in a place that's already obscure due to the many regular expressions, I don't want to make it even more difficult.

Is there any benefit that I'm missing right now?

PS: Btw. you don't need to create an issue for each Pull Request, just put the description into here.

dsidirop:
> Is there any other benefit that I'm missing right now?

Just less maintenance hassle in the sense that you only have to add the new pattern in the patterns-table and enjoy. I will close the PR - maybe adding a comment in that place would help future contributors realize the design intention.

Appreciate the prompt response - be safe and keep rocking!

PR #1518, opened 2025-06-08
unitframes: add raid icon offset and alignment options

clangdo:
I built this because I find the default raid icon placement on unit frames annoying since it can cover text or buffs/debuffs depending on other configuration.

I noticed you removed a lot of offsets in your recent commits, so maybe you don't want this kind of change. Regardless, this allows for customization, but it may be a slightly breaking change in terms of alignment (see TOP->CENTER anchor change).

No need to merge it if you don't like it, but I thought I'd offer. If you'd like stylistic or organizational changes, please just let me know, I'll try to get around to it when I can. Or feel free to restyle it yourself to your liking!

### DIFF
--- a/modules/gui.lua
+++ b/modules/gui.lua
@@ -1046,17 +1046,6 @@ pfUI:RegisterModule("gui", "vanilla:tbc", function ()
         "powermiss:" .. T["Mana - Missing"],
         "powerminmax:" .. T["Mana - Min/Max"],
       },
-      ["uf_raidiconalign"] = {
-        "LEFT:" .. "Left",
-        "RIGHT:" .. "Right",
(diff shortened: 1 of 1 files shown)


shagu:
Looks very good now, thank you! Will test and merge it later today.

shagu:
Thanks, it's merged now.
I decided to slightly update the default values:
https://github.com/shagu/pfUI/commit/860d5cca48dfe3b252531067c62ca02363eb4bce

On default profiles, this looks more similar to how it was before. 
It's not perfect, but I think it should work well for most of the people.

I decided to not write config migrations for it, and just have it like this. 
I hope nobody gets made about the few pixel it may shift on their custom config :see_no_evil:

PR #1519, opened 2025-06-09
Update unitframes.lua

yani9o:
Prevent class colored names to look washed out (especially druid is affected by this and me personally can't differentiate between druids and warriors anymore):

Apply pastel if NOT both are true:
- preset == "unit"
- config["classcolor"] == "1"
- UnitIsPlayer(unitstr)

Before:
image

Now:
image

### DIFF
--- a/api/unitframes.lua
+++ b/api/unitframes.lua
@@ -2602,7 +2602,7 @@ function pfUI.uf.GetColor(self, preset)
   end
 
   -- pastel
-  if C.unitframes.pastel == "1" and not (config["classcolor"] == "1" and UnitIsPlayer(unitstr)) then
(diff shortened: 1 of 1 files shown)


shagu:
Thanks for your contribution, but for me, the top picture looks way better and I'm used to it. 
I'm having an OLED screen with good contrast and color brightness. For me it has been always been very easy to differentiate it, even before the OLED display. 

Just disable `pastel` if your monitor doesn't have the color range/gamma/contrast settings to be able to differentiate them, it's already an option for that reason - I often seen photos of monitors describing the same issue (usually the had incredible overtuned brightness and gamma values)

Sorry, but I won't merge that, as I don't want to play like that, and it would enforce it to everyone else too.
Just disable "pastel" colors if you're screen or eyes are not made for that.

PR #1528, opened 2025-06-14
unitframes: use superwow for clickcasting

francma:
superwow allows us to use 2nd argument of CastSpellByName to specify spell target unit, this is better than the target + targetlast as it doesnt reset autoattack


shagu:
Thanks! Merged :)

PR #1530, opened 2025-06-18
Update autoshift.lua: Moonkin-check wrong number

yani9o:
Autoshift from Moonkin didn't work:
Moonkin Talent is the 15th talent in the tree - 16th is Nature's Grace.
After changing this value, autoshift worked for moonkin form.


shagu:
Thanks for your efforts, but It's a turtle custom thing and I cant merge it like that, as it would break vanilla.

yani9o:
Oh right - sorry.
will go via different approach then.

PR #1533, opened 2025-06-19
WIP: unitframes: new config options regarding the portrait

luskanek:
This PR adds a new option labeled 'Portrait Style' that is available for all unitframes and allows the player to customize the portrait style. Two styles are available for now:
- Default - square style which uses pfUI/Blizzard borders
- Blizzard - circle square which uses the ring texture of the default Blizzard unitframes

Along with this new option comes another new option labeled 'Portrait Offset'. Modifying the offset value will move the portrait along the x-axis.

image
image


shagu:
Thanks for your Pull Request.

For some reason, none of the options made any effect. I selected on my existing config Portrait Style: Blizzard, but nothing changed. Also I made a new character with an empty configuration and had the same result.

If this is a Work-in-Progress and not yet ready to be tested/merged, please flag it in the title as such.

luskanek:
> Thanks for your Pull Request.
> 
> For some reason, none of the options made any effect. I selected on my existing config Portrait Style: Blizzard, but nothing changed. Also I made a new character with an empty configuration and had the same result.
> 
> If this is a Work-in-Progress and not yet ready to be tested/merged, please flag it in the title as such.
> 
> _Edit: I see now it's because of the 2d/3d models. I'm afraid I won't be able to merge it like that. This is very confusing and people will all query me about it. It would require too much time on my end to provide support to it._

Yes it only supports 2D portraits at the moment, however, I'm trying to make it work with 3D portraits too. Until then I will keep this open as WIP.

luskanek:
3D portraits can now be used with the 'Blizzard' option selected. While it does not look perfect in some cases it's the closest to perfect I was able to get (it's a bummer you can't use proper masks).

PR #1561, opened 2025-08-11
Set quest items' bag slot border to the right color under non-English locales

dkeehl:
The item type which GetItemInfo()  returns is a localized name, so under non-English locales just comparing it with string "Quest" will cause the border's color not correctly set. To fix this, add a localized global variable INVTYPE_QUEST and use it in the condition test.


shagu:
Thanks, but this does not work if the eqcompare is disabled. Please don't rely on the global but instead access it via it's locale (`L["itemtypes"]["INVTYPE_QUEST"]`) entry or create a local shortcut to it.

dkeehl:
OK, I added a commit to use the locale entry directly.

PR #1563, opened 2025-08-12
[WIP] add option to fill raid frames top to bottom for horizontal fill

ghost:
As discussed on discord. 
I think the option may need a better name. Does PFui have a way to only show this option if fill is set to horizontal?
Maybe the vertical sorting can be done prettier. I considered messing further with id to do so, but the code becomes pretty hard to read then.


shagu:
Hello, thanks for your contribution. Is this tested and working and doesn't change the default behaviour for people upgrading?

PR #1577, opened 2025-08-17
Add configurable health and power text colors for unit frames

exprecis:
This patch allows users to set custom colors on the value text of health and power bars for all units, along with a base global configuration.


shagu:
Hey, thanks for your contribution. I think this should have more colors, else it would be weird for druids who use most of the different mana types. I suggest a configuration like this into the custom color section (where the bar colors are defined aswell) maybe:

- Use Custom Text Colors (bool)
- Health Text Color (color)
- Mana Text Color (color)
- Rage Text Color (color)
- Energy Text Color (color)
- Focus Text Color (color)

shagu:
Is there a particular reason why you manually refresh all units? Isn't `U[c]` working for you?
That should call the default update function of the unitframes, which also should do what you manually do.
(ref: https://github.com/shagu/pfUI/blob/master/api/unitframes.lua#L425)

exprecis:
Updated the commits to have individually-set colors per power bar type, and then added a second commit to update colors immediately instead of triggering a UI reload. This is my first wow addon contribution, so feel free to massage the commits if it's something you want in the UI but know a bit better how to do it.

PR #1581, opened 2025-08-23
Unify custom power colors

blehz-be:
Power color for text was not the same as for bars. Also the pastel calculation was not uniform.

### DIFF
--- a/api/unitframes.lua
+++ b/api/unitframes.lua
@@ -157,6 +157,31 @@ local function DebuffOnClick()
   end
 end
 
+local function getPowerColor(unitconfig, unitstr)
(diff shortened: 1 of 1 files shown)


shagu:
Thanks, but I did that already a month ago and had to revert it again. 
There were too many people complained about having bad/unreadable colors now.
By that the text on power bars became barely readable because of having the same color.

Here's the revert: https://github.com/shagu/pfUI/commit/28762f85ae393dc2b21d817ae00252709430fed0

I think what it really needs is an option to define custom text colors aswell.
But I didn't had time to review this yet: https://github.com/shagu/pfUI/pull/1577

PR #1593, opened 2025-09-12
Adding colors to indicate day/night in clock

Mats391:
Wanted to go Nightfinsnapper fishing and didnt know when it was day or night. So I added the day/night indication to the clock through some colors. Day is orange, blue is night.
I added no option to enable/disable this. Let me know if this is needed or if colors should be adjusted for contrast etc.

image

image


Mats391:
I also thought about adding a small sun and moon next to it, but couldnt figure out how.

Mats391:
Updated the blue color to have more contrast 
image

PR #1597, opened 2025-09-30
Fix Rake refresh on Ferocious bite

Aalandriel:
Applied the same change for Rake as was already in place for Rip. Seems to work correctly or at least gives similar result to dot tracking with cursive.


Aalandriel:
Gonna go the edit route to not show the whole file as changed

PR #1602, opened 2026-01-07
fix(cooldown): cache frame reference to reduce GC pressure

ncerny:
## Summary
- Cache cooldown frame reference when `pfCooldownText` is created instead of building string every frame
- Eliminates ~6000+ string allocations/second from the cooldown module
- String concatenation `parent_name .. "Cooldown"` was running every frame before throttle check

## Test plan
- [x] Verify cooldowns display correctly on action bars
- [x] Verify cooldowns hide when underlying animation hides
- [x] Monitor memory allocation rate with performance addon

### DIFF
--- a/modules/cooldown.lua
+++ b/modules/cooldown.lua
@@ -6,17 +6,15 @@ pfUI:RegisterModule("cooldown", "vanilla:tbc", function ()
   local hourcolor   = {strsplit(",", C.appearance.cd.hourcolor)}
   local daycolor    = {strsplit(",", C.appearance.cd.daycolor)}
 
-  local parent, parent_name
+  local parent
   local function pfCooldownOnUpdate()
     parent = this:GetParent()
-    if not parent then this:Hide() end
-    parent_name = parent:GetName()
+    if not parent then this:Hide() return end
 
-    -- avoid to set cooldowns on invalid frames
-    if parent_name and _G[parent_name .. "Cooldown"] then
-      if not _G[parent_name .. "Cooldown"]:IsShown() then
-        this:Hide()
-      end
+    -- avoid to set cooldowns on invalid frames (use cached reference)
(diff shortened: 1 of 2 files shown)


ncerny:
## Issue

The `pfCooldownOnUpdate` function performs string concatenation **before** the throttle check, causing excessive garbage collection pressure.

### Problem Code (lines 10-23)

```lua
local function pfCooldownOnUpdate()
  parent = this:GetParent()
  if not parent then this:Hide() end
  parent_name = parent:GetName()

  -- These run EVERY FRAME before throttle check on line 23
  if parent_name and _G[parent_name .. "Cooldown"] then      -- STRING ALLOC #1
    if not _G[parent_name .. "Cooldown"]:IsShown() then      -- STRING ALLOC #2
      this:Hide()
    end
  end

  -- only run every 0.1 seconds from here on
  if ( this.tick or .1) > GetTime() then return else this.tick = GetTime() + .1 end
```

### Impact

- String concatenation `parent_name .. "Cooldown"` runs **every frame** (60+ times/second)
- Happens **twice** per frame (lines 16 and 17)
- Multiplied by number of active cooldowns (50+ action buttons, inventory items, etc.)
- **Result:** 6000+ temporary string allocations per second

## Solution

Cache the cooldown frame reference when `pfCooldownText` is created, and use the cached reference in OnUpdate instead of rebuilding the string each frame.

```lua
-- In pfCreateCoolDown, cache the reference once:
local parent_name = cooldown:GetParent() and cooldown:GetParent():GetName()
if parent_name then
  cooldown.pfCooldownText.cooldownRef = _G[parent_name .. "Cooldown"]
end

-- In pfCooldownOnUpdate, use cached reference:
if this.cooldownRef and not this.cooldownRef:IsShown() then
  this:Hide()
  return
end
```

PR #1604, opened 2026-01-07
fix(castbar): throttle timer text updates to reduce GC pressure

ncerny:
## Summary
- Only update castbar timer text when displayed value actually changes
- Track displayed value as integer tenths (`floor(cur * 10)`)
- Cache max duration string since it doesn't change during cast
- Reduces string allocations from 60+/second to ~10/second per castbar

## Test plan
- [x] Verify castbar timer text updates smoothly during casting
- [x] Verify delay indicator shows correctly when cast is pushed back
- [x] Verify timer resets properly between casts
- [ ] Monitor memory allocation rate with performance addon


ncerny:
## Issue

The castbar OnUpdate handler performs string concatenation and `string.format()` operations every frame while casting, even when the displayed value hasn't changed.

### Problem Code (lines 150-157)

```lua
if this.showtimer then
  if this.delay and this.delay > 0 then
    local delay = "|cffffaaaa" .. (channel and "-" or "+") .. round(this.delay,1) .. " |r "
    this.bar.right:SetText(delay .. string.format("%.1f",cur) .. " / " .. round(max,1))
  else
    this.bar.right:SetText(string.format("%.1f",cur) .. " / " .. round(max,1))
  end
end
```

### Impact

- `string.format("%.1f", cur)` creates a new string every frame
- String concatenation `..` creates multiple temporary strings every frame
- 60+ frames/second during casting = 60+ string allocations/second per castbar
- With player + target + focus castbars, this multiplies

## Solution

Only update the timer text when the displayed value changes. Since we display with 1 decimal place, track when `floor(cur * 10)` changes:

```lua
if this.showtimer then
  local curTenths = math.floor(cur * 10)
  if this.lastCurTenths ~= curTenths or this.lastDelay ~= this.delay then
    this.lastCurTenths = curTenths
    this.lastDelay = this.delay

    -- cache max string since it doesn't change during cast
    if not this.maxStr then
      this.maxStr = round(max, 1)
    end

    -- only now do string operations
    if this.delay and this.delay > 0 then
      ...
    end
  end
end
```

This reduces updates from 60+/second to ~10/second (once per tenth of a second change).

PR #1606, opened 2026-01-07
fix(api): cache GetColoredTimeString results to reduce GC pressure

ncerny:
## Summary
- Cache formatted time strings keyed by rounded time value
- Repeated calls for same time value return cached string
- Separate caches for days, hours, minutes, seconds, and milliseconds
- Most cooldowns show common values (5,4,3,2,1) that hit cache repeatedly

## Test plan
- [ ] Verify cooldown text displays correctly for all time ranges
- [ ] Verify days/hours/minutes/seconds/milliseconds all format correctly
- [ ] Monitor memory allocation with many active cooldowns


ncerny:
## Issue

The `GetColoredTimeString` function performs string concatenation on every call, creating garbage. This function is called every 0.1 seconds for each active cooldown.

### Problem Code (lines 1219-1279)

```lua
function pfUI.api.GetColoredTimeString(remaining)
  -- ...
  return color_low .. round(remaining)  -- String allocation every call
  -- ...
  return color_normal .. round(remaining)  -- String allocation every call
end
```

### Impact

- Called every 0.1 seconds per active cooldown
- With 10+ cooldowns active, that's 100+ string allocations per second
- Cooldowns often show the same values (5, 4, 3, 2, 1 seconds)
- Same strings are recreated repeatedly instead of reused

## Solution

Cache formatted strings for each unique time value. Since cooldowns frequently show the same integer/decimal values, cache these pre-formatted strings:

```lua
local time_cache_low = {}      -- integer seconds 0-5
local time_cache_normal = {}   -- integer seconds 6-99
local time_cache_ms = {}       -- milliseconds 0.0-5.0
local time_cache_minute = {}   -- minutes
local time_cache_hour = {}     -- hours
local time_cache_day = {}      -- days

-- Example for normal seconds:
local rounded = round(remaining)
if not time_cache_normal[rounded] then
  time_cache_normal[rounded] = color_normal .. rounded
end
return time_cache_normal[rounded]
```

Now repeated calls for "5", "4", "3", "2", "1" return cached strings instead of allocating new ones.

PR #1607, opened 2026-01-07
fix(buff): use table.concat to avoid O(n²) string allocations

ncerny:
## Summary
- Replace string concatenation loop with `table.insert()` + `table.concat()`
- Fixes O(n²) memory allocation pattern in "Unbuffed" tooltip list
- In 40-person raid, each Shift-hover was creating 40+ intermediate strings
- Fixed in both `modules/buff.lua` and `api/unitframes.lua` (duplicate code)

## Test plan
- [ ] Verify buff tooltip "Unbuffed" list shows correctly in raid
- [ ] Verify buff tooltip works correctly in party  
- [ ] Verify player names display with correct colors
- [ ] Monitor memory allocation when hovering buffs with Shift


ncerny:
## Issue

The buff tooltip "Unbuffed" list feature uses string concatenation in a loop, causing O(n²) memory allocations.

### Problem Code (buff.lua lines 155-179, unitframes.lua lines 76-100)

```lua
local playerlist = ""
local first = true

if UnitInRaid("player") then
  for i=1,40 do
    local unitstr = "raid" .. i
    if not UnitHasBuff(unitstr, texture) and UnitName(unitstr) then
      -- O(n²) pattern: each concat creates a new string copying all previous content
      playerlist = playerlist .. ( not first and ", " or "") .. GetUnitColor(unitstr) .. UnitName(unitstr) .. "|r"
      first = nil
    end
  end
end
```

### Impact

- Each `playerlist = playerlist .. ...` creates a new string
- In a 40-person raid, this can create 40+ intermediate strings
- O(n²) memory usage pattern (total bytes allocated grows quadratically with list size)
- Happens on every buff tooltip hover with Shift held
- Same issue exists in both `modules/buff.lua` and `api/unitframes.lua`

## Solution

Use `table.insert()` to collect player names, then `table.concat()` once at the end:

```lua
local playerlist = {}

if UnitInRaid("player") then
  for i=1,40 do
    local unitstr = "raid" .. i
    if not UnitHasBuff(unitstr, texture) and UnitName(unitstr) then
      table.insert(playerlist, GetUnitColor(unitstr) .. UnitName(unitstr) .. "|r")
    end
  end
end

if table.getn(playerlist) > 0 then
  GameTooltip:AddLine(table.concat(playerlist, ", "),1,1,1,1)
end
```

This changes from O(n²) to O(n) memory allocation.

PR #1609, opened 2026-02-15
/castfocus, /pfcastfocus: Optimize castfocus using superwow; also side-optimize and enhance /pfcast to employ memoization under the hood and support passing raw-lua-functions to it

dsidirop:
- Add support for superwow-aware SlashCmdList.PFCASTFOCUS() just like we did with SlashCmdList.PFFOCUS()

This allows us to employ unit-guid-targeting to side-step the complex target-swapping tricks we would resort to when superwow wasn't available.
   
This is a much more care-free and lightweight approach.

- Replace loadstring() calls in _G.SlashCmdList.PFCAST() and SlashCmdList.PFCASTFOCUS() with the new pfUI.api.TryMemoizedFuncLoadstringForSpellCasts() to enjoy better runtime performance when the user passes lua-func-strings (considering that they're the same strings time and over again).

As an added bonus this commit also adds seamless support for passing actual raw functions to SlashCmdList.PFCASTFOCUS() ala:

```lua
SlashCmdList.PFCASTFOCUS(function() CastspellByName("Renew" .. desiredRank); end)
``` 

This is extremely useful for advanced pure-lua macros (p.e. via SuperMacro) as it simplifies the invocation-style tremendously.

Old was: Before this commit we would have to write this kind of stuff as a string

```lua
SlashCmdList.PFCASTFOCUS( 'function() CastspellByName("Renew"' .. desiredRank .. '); end' )  -- notice we had to use a "string-function" here!
```

This was both 10x slower and was extremely cumbersome and error-prone for the user if he wanted to pass some variable-state inside the string-func-callback (god help him in that case!)


dsidirop:
This PR was based on the wrong branch. Closed it and re-opened it using the correct branch in #1610 so have a look at that one.

